PR #6, opened 2017-07-14
MAGECLOUD-872 Changed executeDbQuery to use mysqli instead of command line execution of mysql client

JacobBrownAustin:
MAGECLOUD-872 Changed executeDbQuery to use mysqli instead of command line execution of mysql client

Create new Database class to help make this clean and easier if we need to add more features to this.
Now we avoid shell/command injection bugs/exploits by using mysqli instead of execute.
Now we avoid SQL injection by using parameterized queries.
Avoiding exposing password to command line arguments.
Better error detection and error messages with the exception throws I added to the new database class.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 .idea
+vendor
+composer.phar
+phpunit.xml
+composer.lock
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 vendor
 composer.phar
 phpunit.xml
-composer.lock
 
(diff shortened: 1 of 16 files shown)


ngeron (commented):
Looks good. Tested with 2 deployments.

mazhalai (dismissed):
Also, there is one change missed in src/Magento/MagentoCloud/Command/SCDConfigDump.php which will also take care of removal of duplication of code task MAGECLOUD-772

PR #22, opened 2017-08-21
Magecloud-894 Remove default admin username/password and send password reset email on install to ADMIN_EMAIL

JacobBrownAustin:
https://magento2.atlassian.net/browse/MAGECLOUD-894
and 
https://magento2.atlassian.net/browse/MAGECLOUD-115

We used to use a default admin username/password when ADMIN_USERNAME and ADMIN_PASSWORD weren't set.  This was bad, because the potential for people to forget to change them.

Now, we do different things based on whether it is installing or not.

On install, if ADMIN_USERNAME and ADMIN_PASSWORD are not set, the ADMIN_PASSWORD is set to a randomly generated password, and an email to reset the password is sent to ADMIN_EMAIL.  If ADMIN_EMAIL is also not set, an exception is thrown at the start of Deploy (which can be seen in /var/log/deploy.log) , and website is put in maintenance mode.  Note that ADMIN_EMAIL is being added to the cloud onboarding process, so it should always be set.  By default, "master" branch of the project is set to enable email sending, so the reset email is sent to the ADMIN_EMAIL address, and they can reset the password there.  Because the reset password email template does not include the username of the admin, we still have to use a default of "admin" for ADMIN_USERNAME.  That will have to be fixed in a later issue/pull request.

When deploy is run on an already existing site (ie: not installing), no default values are loaded into the variables.  I have modified updateAdminCredentials() method so that only non-empty variables are updated.  Variables that are empty are no longer updated in that method.

I've added a "m2-ece-send-password-reset-email" command that sends the email.  Eventually, I would like this command (and other similar ones) to be part of core Magento so that we don't have to maintain it here.  It would also be useful to non-cloud customers.

In installMagento() method, I had to add escapeshellarg calls to get around some errors that were happening with some of the random passwords.  I've been wanting to add escapeshellarg calls there anyways for security reasons, but this issue required it there.

I fixed two small security things with generatePassword() method.

I did add a few TODOs to the code.  Hopefull someone can comment on them:
1.  In m2-ece-send-password-reset-email command, I had a couple of issues with bootstrapping.  I found require_once files that got it to work, but it doesn't look like the correct way.  Please review.

2. In SendPasswordResetEmail class, I'm using hard-coded user_id = 1.   Hard-coding values like this seems bad, but this is also used in Deploy for the admin user.   Can someone confirm that this is okay?

3.  I found an issue with sending emails from command line where the command name was being prepended to the path of the URL as seen in the emails being sent to the users.  I found a workaround by setting SCRIPT_FILENAME to "", but this seems wrong.  I will probably have to write this up as a core Magento issue.


shiftedreality (dismissed):
This solution must be discussed with architect @kandy  nad @NadiyaS

Also forward-port is required in any case

kandy (changes requested):

kandy (approved):

mazhalai (approved):

PR #41, opened 2017-09-25
MAGECLOUD-1020: [develop] Enterprise Cloud Edition Sets Up an Admin User with A Default Password

BaDos:
### Description

*In install process:*
1. We always require ADMIN_EMAIL
2. If ADMIN_PASSWORD is not set we send email for reset password and save mail to var/credentials_email.txt, else not send
3. By default, we use admin URL is admin and admin username is admin
4. If ADMIN_USERNAME is not set we use its value as admin username
5. If ADMIN_URL is not set we use its value as admin URL

*In update process:*
0. All admin information is updated only for first admin user
1. We don't require ADMIN_EMAIL like required variable
2. If ADMIN_EMAIL is set we check that nobody uses this email
2.1 If nobody uses we update email address for first admin, else log message that somebody uses this email address
3. If ADMIN_USERNAME is set we check that nobody uses this username
3.1 If nobody uses we update username for first admin, else log message that somebody uses this username

https://magento2.atlassian.net/browse/MAGECLOUD-1020

### Manual testing scenarios

https://jira.corp.magento.com/browse/MAGETWO-80875

### Contribution checklist
 - [ ] Pull request was approved by architect


JacobBrownAustin (changes requested):
Since it won't let me comment on lines that haven't been edited,
https://github.com/magento/ece-tools/pull/41/commits/d8a7e0edc80262f7066281343239ef5935eb5e5a#diff-dd69dda0c05e68307f4afe9db72ea47fR315
On this line, since you are hard-coding the default value for getAdminUsername(), this causing a user who set their username via the normal Magento way to have ther username overwritten back to the default when a redeploy happens.
Example flow:
The user never sets ADMIN_USERNAME
They Install Magento in our cloud environment.
The username is now 'admin'.
User now changes their username to 'root' via normal Magento interface;
The username is now 'root'.
A new version of Magento is released, they decide to upgrade, redeploy happens.
During redeploy,  Magento\MagentoCloud\Process\Deploy\InstallUpdate\Update::execute() calls getAdminUsername().
Username is now 'admin'.
Same issue for Environment::getAdminUrl() when it is called by EnvConfiguration::execute() .

shiftedreality (changes requested):

shiftedreality (approved):

PR #72, opened 2017-10-31
MAGECLOUD-1122:  Fixing cron:run on non-production environments.

JacobBrownAustin:
https://magento2.atlassian.net/browse/MAGECLOUD-1122
### Description
We need to check if we are currently in a non-production environment, and if we are, then modify the module-cron so that it works around the fact that our cron:run will only run once every five minutes.  To do this, this pull request checks the environment variables MAGENTO_CLOUD_PROJECT and HOME to see if it is a production environment.  If it isn't, then it sets the $config["system"]["default"]["cron"]["cron_interval"] = "5"; in env.php .
This variable is used later in a patch that has a separate pull request in ece-patches.
It would be better if we could determine the cron interval on the current environment directly, but I asked Platform.sh, and there is no way to currently do this, so MAGENTO_CLOUD_PROJECT and HOME were the best that I could do for now.

1. Find a way to determine whether or not we are in a production environment.
2. Set a variable in env.php that is later used by a patch in magento-patches

### Zephyr Tests

1. ...

### Contribution checklist
 - [ ] Pull request was approved by architect
 - [ ] Pull request was approved by QA member


shiftedreality (changes requested):
We should eliminate patch and think about more elegant solution, maybe using `config:set --lock`.
@JacobBrownAustin please book a meeting to discuss this item.

shiftedreality (changes requested):

billygilbert (changes requested):

michaelcasey316 (approved):

PR #160, opened 2018-01-24
Revert "MAGECLOUD-1086 Migrate Magento 2.1.x versions to ECE-Tools & Patches"

billygilbert:
Reverting Merge so that we can test in stand alone branch.


JacobBrownAustin (approved):
Okay.  We revert for now.

PR #324, opened 2018-07-30
MAGECLOUD-2521: Patch for Zendframework1.  Use TLS 1.2

JacobBrownAustin:

1. https://magento2.atlassian.net/browse/MAGECLOUD-2521

### Manual testing scenarios

- https://jira.corp.magento.com/browse/MAGETWO-94258

Because I couldn't find an open or free TLS testing service, to test and verify this patch against SMTP, I made this phpunit file that can run to verify the functionality of this patch.  

All it does is try to Spark Post Mail's SMTP server, which doesn't support TLS 1.0, so it requires this patch.  It then tries to send an email, but fails, because I didn't want to put my real test account credentials in there.  So the test verifies that it fails because of Authentication error instead of failing because of TLS error.

To run it, download this file and add it to the root of your cloud environment. 
https://github.com/magento-cloud/magento-cloud-misc/blob/master/users/jacob/MAGECLOUD-2521/test-smtp-tls-sparkmail.php
Also, make sure to add php unit to your cloud environment by running this command.
`composer require phpunit/phpunit`
Then add it all and push.
`git add composer.lock composer.json test-smtp-tls-sparkmail.php ; git commit -m 'adding test for SMTP TLS' ; git push`
Once your cloud environment had redeployed, ssh into it and run 
`phpunit test-smtp-tls-sparkmail.php`
If you have the patch applied , it will succeed without failures.  If you don't have this patch applied, it will fail.

Note that this test will not work with PHP 7.2, because in PHP 7.2, they switched the TLS constants again, so it will silently connect with TLS 1.2 when using PHP 7.2.

Note that this test only tests SMTP and does not test the other affected protocols, POP3, IMAP, and HTTP proxy adaptor.

### Contribution checklist


shiftedreality (changes requested):

shiftedreality (approved):

shiftedreality (changes requested):

billygilbert (approved):

YPyltiai (changes requested):
Align patch format with all other patches, please.

YPyltiai (approved):
Need to find some good way to test it now. Everything else looks good. Thanks for fixing @shiftedreality

shiftedreality (approved):

PR #399, opened 2018-12-17
MAGECLOUD-2904: SCD_EXCLUDE_THEMES / SCD_MATRIX - Resolve Issues with Case-sensitive theme name

billygilbert:
### Description

Added case validation to SCD_MATRIX and SCD_EXCLUDE_THEMES.  Both processes will check all themes passed in against an array of themes in vendor as well as app/design.  If the theme exists with the same name but different case, the case will be corrected to what is defined in registration.php.  If it is not found, it will be unset and a warning will be thrown.

1.  MAGECLOUD-2904

### Manual testing scenarios

1. Updated Zephyr https://jira.corp.magento.com/browse/MAGETWO-71869
2. Updated Zephyr https://jira.corp.magento.com/browse/MAGETWO-90874

### Contribution checklist

### DIFF
--- a/src/Filesystem/DirectoryList.php
+++ b/src/Filesystem/DirectoryList.php
@@ -27,11 +27,9 @@ class DirectoryList
     const DIR_GENERATED_CODE = 'code';
     const DIR_GENERATED_METADATA = 'metadata';
     const DIR_ETC = 'etc';
-    const DIR_DESIGN = 'design';
     const DIR_MEDIA = 'media';
     const DIR_VIEW_PREPROCESSED = 'view-preprocessed';
     const DIR_STATIC = 'static';
-    const DIR_VENDOR = 'vendor';
 
     /**
      * @var string
@@ -187,11 +185,9 @@ class DirectoryList
             static::DIR_VAR => [static::PATH => 'var'],
             static::DIR_LOG => [static::PATH => 'var/log'],
             static::DIR_ETC => [static::PATH => 'app/etc'],
-            static::DIR_DESIGN => [static::PATH => 'app/design'],
             static::DIR_MEDIA => [static::PATH => 'pub/media'],
             static::DIR_STATIC => [static::PATH => 'pub/static'],
             static::DIR_VIEW_PREPROCESSED => [static::PATH => 'var/view_preprocessed'],
-            static::DIR_VENDOR => [static::PATH => 'vendor'],
         ];
 
         return $config;
--- a/src/Filesystem/DirectoryList.php
+++ b/src/Filesystem/DirectoryList.php
@@ -27,9 +27,11 @@ class DirectoryList
     const DIR_GENERATED_CODE = 'code';
     const DIR_GENERATED_METADATA = 'metadata';
     const DIR_ETC = 'etc';
+    const DIR_DESIGN = 'design';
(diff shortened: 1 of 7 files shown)


src/StaticContent/ThemeResolver.php
```diff
@@ -0,0 +1,116 @@
...
+        });
+
+        return $themes;
+    }
+
+    /**
+     * @param string $themePath
+     * @return string
+     */
+    public function getThemeName(string $themePath): string
```
bbatsche: Man I don't like this method, but I don't have a good answer for what the alternative should be

src/Filesystem/Driver/File.php
```diff
@@ -477,4 +477,9 @@ public function requireFile(string $path)
     {
         return require $path;
     }
+
+    public function glob(string $path): array
```
shiftedreality: It's better to pass `NO_SORT` option by default, this will improve performance

src/StaticContent/ThemeResolver.php
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Parses the registration file of the theme to get the name of the theme as registered.
+     * @param string $themePath
+     * @return string
+     */
+    public function getThemeName(string $themePath): string
+    {
+        try {
```
shiftedreality: I see you already adding coupling on Magento. So maybe it's better to retrieve them from `\Magento\Framework\Component\ComponentRegistrar:$paths`?

I see it's private but we can use reflection easily:

```
if (class_exists(\Magento\Framework\Component\ComponentRegistrar::class)) {
    $reflectionClass = new \ReflectionClass(\Magento\Framework\Component\ComponentRegistrar::class);
    $property = $reflectionClass->getProperty('paths');
    $property->setAccessible(true);

    $paths = $property->getValue($reflectionClass)[\Magento\Framework\Component\ComponentRegistrar::THEME];

    die(var_dump(array_keys($paths)));
}
```

src/Filesystem/DirectoryList.php
```diff
@@ -27,9 +27,11 @@ class DirectoryList
     const DIR_GENERATED_CODE = 'code';
     const DIR_GENERATED_METADATA = 'metadata';
     const DIR_ETC = 'etc';
+    const DIR_DESIGN = 'design';
```
shiftedreality: It seems like we can remove both of these constants

src/StaticContent/ThemeResolver.php
```diff
@@ -0,0 +1,105 @@
...
+                array_map('strtolower', $availableThemes)
+            );
+            if (false !== $themeNamePosition) {
+                $this->logger->warning(
+                    'Theme found as ' . $availableThemes[$themeNamePosition] . '.  Using corrected name instead.'
+                );
+                return $availableThemes[$themeNamePosition];
+            }
+        }
+        return '';
```
shiftedreality: Can we have a fallback when no theme foud and use provided value? Otherwise, it will break local setup having ece-tools dependent on Magento

shiftedreality (changes requested):

shiftedreality (approved):

andriyShevtsov:
QA  approved

PR #389, opened 2018-11-19
MAGECLOUD-2956: Exception during build phase: Undefined class constant 'PARSE_CONSTANT'

oshmyheliuk:
### Fixed Issues (if relevant)

https://magento2.atlassian.net/browse/MAGECLOUD-2956

### Manual testing scenarios

https://jira.corp.magento.com/browse/MAGETWO-95466

### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #404, opened 2019-01-07
Bad constraint "symfony/yaml": "^2.3||^3.3"

vkerkhoff:
Bad constraint `"symfony/yaml": "^2.3||^3.3"`

### Description
vendor/magento/ece-tools/src/Config/Environment/Reader.php on line 58 uses Yaml::PARSE_CONSTANT, but this constant is only introduced in Symphony Yaml v3.2
With the current constraints, composer allows us to run Yaml v2.8 due to a different module requiring "^2.0", but the build breaks on `vendor/bin/m2-ece-build`

Solutions it to remove ^2.3 from the ece-tools constraint.

### Contribution checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "psr/container": "^1.0",
     "psr/log": "^1.0",
     "symfony/console": "^2.6||^4.0",
-    "symfony/yaml": "^2.3||^3.3",
+    "symfony/yaml": "^3.3",
     "twig/twig": "^1.0||^2.0"
   },
   "require-dev": {


shiftedreality:
@vkerkhoff Thanks for your contribution!

Unfortunately, this change will break compatibility with n98-magerun2 utility https://github.com/netz98/n98-magerun2/blob/develop/composer.json#L33

vkerkhoff:
But currently, it breaks the build if it is installed with symfony/yaml <3.2, I would rather receive an error message during composer install that it has a conflict. Also, most of the time you will install the phar of magerun2 and not include it in your project, so that won't cause a conflict then. And most important, ece-tools isn't compatible with symfony/yaml <3.2 so it shouldn't be listed as a dependency.

YPyltiai:
@vkerkhoff actually https://github.com/magento/ece-tools/pull/389 should have fixed this. It will be in .16 release (.15 now has the bug mentioned).

shiftedreality:
@vkerkhoff please check Yevhenii's comment. Is it still actual after #389 was delivered?

vkerkhoff:
@shiftedreality I guess it solves the problem, but not sure if this is really a clean solution as it can give unpredictive results (with >3.2 installed it might give a different result compared to <3.2). I think that just removing the 2.3 dependencies would be better as my previous comment.

YPyltiai:
@cmuench , would it be possible to add ^3.3 support to https://github.com/netz98/n98-magerun2 ? I was pinging you on Slack previously but didn't remember getting a response.

YPyltiai:
Opened an issue: https://github.com/netz98/n98-magerun2/issues/430

YPyltiai:
Based on the information from https://github.com/netz98/n98-magerun2/issues/430 - we further agree that this utility should be used in form of `phar` as per official recommendations. Outdated `symfony/yaml` version can be removed from ECE-Tools dependencies.

PR #407, opened 2019-01-11
MAGECLOUD-3048: Allow setting of x-frame options within configuration…

shiftedreality:
… file

1. https://magento2.atlassian.net/browse/MAGECLOUD-3048

### Manual testing scenarios

1. https://jira.corp.magento.com/browse/MAGETWO-97626

### Contribution checklist

### DIFF
--- a/src/Config/Schema.php
+++ b/src/Config/Schema.php
@@ -406,6 +406,15 @@ class Schema
                     StageConfigInterface::STAGE_POST_DEPLOY => [''],
                 ],
             ],
+            StageConfigInterface::VAR_X_FRAME_CONFIGURATION => [
+                self::SCHEMA_TYPE => ['string'],
+                self::SCHEMA_STAGE => [
+                    StageConfigInterface::STAGE_GLOBAL
+                ],
+                self::SCHEMA_DEFAULT_VALUE => [
+                    StageConfigInterface::STAGE_GLOBAL => ''
+                ]
+            ]
         ];
     }
 }
--- a/src/Config/Schema.php
+++ b/src/Config/Schema.php
@@ -30,7 +30,7 @@ class Schema
      * @param string $stage
      * @return array
      */
-    public function getDefaults(string $stage)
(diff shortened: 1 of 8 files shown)


src/Config/Schema.php
```diff
@@ -406,6 +406,15 @@ public function getSchema()
                     StageConfigInterface::STAGE_POST_DEPLOY => [''],
                 ],
             ],
+            StageConfigInterface::VAR_X_FRAME_CONFIGURATION => [
+                self::SCHEMA_TYPE => ['string'],
+                self::SCHEMA_STAGE => [
+                    StageConfigInterface::STAGE_GLOBAL
+                ],
+                self::SCHEMA_DEFAULT_VALUE => [
+                    StageConfigInterface::STAGE_GLOBAL => ''
```
billygilbert: looks like default should be SAMEORIGIN.  That's what I get in my env.php without changing anything.

billygilbert (changes requested):
Also needs an update to dist/.magento.env.yaml

src/Config/Schema.php
shiftedreality: This will be backward incompatible because will overwrite custom values.

YPyltiai:
@shiftedreality / @billygilbert - would it be better to move it into `system` section? Doesn't logically fit into `global` (no effect on deployment process).  It's not a requirement - just a question.

shiftedreality:
@billygilbert fix both comments, changing this value by default seems not very risky.
@YPyltiai it's on the same level as cache settings for example

YPyltiai:
Then it's fine.

billygilbert:
Maybe at some point it makes sense to move these all to an `SYSTEM>ENV` section...or something that denotes it specifically affects env.php but I agree with @shiftedreality for now.  Approved.

billygilbert (approved):

mveeramneni:
QA Approved

PR #416, opened 2019-02-08
MAGECLOUD-2793: Patch Monolog 1.6.x to Support ECE-Tools Features for Magento 2.1

oshmyheliuk:
### Description
Magento 2.1 is locked to using Monolog 1.6. Monolog 1.6 has a bug (resolved in 1.18.1) where the vast majority of Slack messages are not being sent.

https://magento2.atlassian.net/browse/MAGECLOUD-2793

### Manual testing scenarios

1. Configure Slack Handler in `.magento.env.yaml`
```
log:
  slack:
      token: "xoxp-xxxx"   #your token here
      channel: "xxxxx"     # channel name
      username: "SlackHandler"
      min_level: "info"
```
2. Redeploy magento 2.1.x on cloud
3. Check that deployment log messages were send to your Slack chanel from step 1.

### Contribution checklist

### DIFF
--- a/patches.json
+++ b/patches.json
@@ -138,6 +138,9 @@
     "Add the possibility to install Magento without admin creation" : {
       "2.1.4 - 2.2.1": "MAGECLOUD-2573__installation_without_admin_creation__2.1.4.patch",
       "2.2.2 - 2.2.7": "MAGECLOUD-2573__installation_without_admin_creation__2.2.2.patch"
+    },
+    "Fix monolog Slack Handler bug for magento 2.1.x": {
+      "~2.1.4": "MAGECLOUD-2793__fix_monolog_slack_handler_2.1.x.patch"
     }
   }
 }
--- a/patches.json
+++ b/patches.json
@@ -138,9 +138,11 @@
     "Add the possibility to install Magento without admin creation" : {
       "2.1.4 - 2.2.1": "MAGECLOUD-2573__installation_without_admin_creation__2.1.4.patch",
       "2.2.2 - 2.2.7": "MAGECLOUD-2573__installation_without_admin_creation__2.2.2.patch"
-    },
+    }
+  },
+  "monolog/monolog": {
     "Fix monolog Slack Handler bug for magento 2.1.x": {
-      "~2.1.4": "MAGECLOUD-2793__fix_monolog_slack_handler_2.1.x.patch"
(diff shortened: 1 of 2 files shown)


patches.json
```diff
@@ -138,6 +138,9 @@
     "Add the possibility to install Magento without admin creation" : {
       "2.1.4 - 2.2.1": "MAGECLOUD-2573__installation_without_admin_creation__2.1.4.patch",
       "2.2.2 - 2.2.7": "MAGECLOUD-2573__installation_without_admin_creation__2.2.2.patch"
+    },
+    "Fix monolog Slack Handler bug for magento 2.1.x": {
+      "~2.1.4": "MAGECLOUD-2793__fix_monolog_slack_handler_2.1.x.patch"
```
shiftedreality: You should depend on the `monolog/monolog` package rather Magento

shiftedreality (changes requested):

patches.json
oshmyheliuk: As I remember dependencies on modules were removed in this task https://magento2.atlassian.net/browse/MAGECLOUD-1648

```
To do:
..
Change patch constraints so they're based on Magento version and not individual module versions.
```

YPyltiai: Correct. Since Magento versions have hard-coded version of libraries, we depend on core version for the mapping to be more straightforward and easy to read. Same approach was taken for `credis` previously. So this looks to be OK as soon as constraint is closed. Why do we apply it to `~2.1.4` - which version is used on latest `2.1.x` branch?

YPyltiai: Version of lib is not likely to change. Just confirmed with patch team:
`correct, we do only security fixes for 2.1.x, so external library can be updated only in cases when the current version is no longer supported or missing some security fixes`

We should be safe to use core version here but can use module version as an exception here, since constraint is still open and there is some risk that dependency will break if eventually monolog requires a security fix.

oshmyheliuk: All Magento 2.1.x versions are using "monolog/monolog": "1.16.0". So, I applied this patch to "~2.1.4".
Since Magento 2.2.0 constraint was changed to "monolog/monolog": "^1.17".

oshmyheliuk: I will change dependency on "monolog/monolog".

shiftedreality (approved):

andriyShevtsov:
@oshmyheliuk  please add this test to description https://jira.corp.magento.com/browse/MAGECLOUD-39

andriyShevtsov:
QA passed

PR #431, opened 2019-03-14
MAGECLOUD-3275: Rewrite ConfigDumpTest to use Cloud Docker

shiftedreality:

1. https://magento2.atlassian.net/browse/MAGECLOUD-3275

### Contribution checklist


src/Test/DockerIntegration/ConfigDumpTest.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    public function test()
+    {
+        (new Process\GitClone('master'))
+            ->setTimeout(null)
+            ->mustRun();
+        (new Process\ComposerInstall())
+            ->setTimeout(null)
+            ->mustRun();
+
```
billygilbert: Should we look to start abstracting some of these out to helpers?  For instance, the above 2 sets of code seem to be called in every test.  I'm also thinking we could combine some of these tests for speed (though I can't quite see the design in my head right now) but for example, this test and AcceptanceTest are the exact same from lines 18-41 so maybe those could, at a minimum, be moved to helper functions, but even possibly be combined so they are a single deploy process

billygilbert (commented):
I don't have any required changes, just some long term strategic thoughts with what we're doing here.  If you think it's best to review that in a separate ticket or have an offline conversation, I approve the PR.

src/Test/DockerIntegration/ConfigDumpTest.php
shiftedreality: Merged ConfigDumTest into AcceptanceTest

src/Test/DockerIntegration/AcceptanceTest.php
```diff
@@ -41,11 +47,44 @@ public function testDefault()
...
+        $read = new Process\Copy('/app/etc/config.php', $to);
+        $read->mustRun();
+
+        $config = require $to;
+
+        $arrayManager = new ArrayManager();
+        $flattenKeysConfig = implode(array_keys($arrayManager->flatten($config, '#')));
+
+        $this->assertContains('#modules', $flattenKeysConfig);
+        $this->assertContains('#scopes', $flattenKeysConfig);
```
billygilbert, line 77: One last though, is it helpful to add the third parameter $message to these so that if there is a failure, it's a bit easier to troubleshoot?

shiftedreality: It's pretty self-explaining the kets are different to assert and will be displayed in case of fail

billygilbert (approved):

billygilbert (approved):

PR #443, opened 2019-04-15
MAGECLOUD-3369: [Cloud Docker] Redis is Required for Deploy Phase even if it is Absent in Services

oshmyheliuk:
### Description

Deploy failing after removing Redis from services.

https://magento2.atlassian.net/browse/MAGECLOUD-3369

### Manual testing scenarios

https://jira.corp.magento.com/browse/MAGECLOUD-111

### Contribution checklist


src/Docker/Config/Relationship.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Generates relationship data for current configuration
+     *
+     * @throws ConfigurationMismatchException
+     */
+    public function get(): array
+    {
+        $magentoCloudRelationships = [];
```
shiftedreality: Let's keep naming simpler, `$rlationships` would be fine

src/Docker/Config/RelationshipUpdater.php
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Updates MAGENTO_CLOUD_RELATIONSHIPS section in docker config.php.* files according to services enablements.
+     */
+    public function update()
+    {
+        $magentoRoot = $this->directoryList->getMagentoRoot();
+
+        $files = [
+            $magentoRoot . '/docker/config.php',
```
shiftedreality: We should not change custom file

src/Docker/Config/RelationshipUpdater.php
```diff
@@ -0,0 +1,124 @@
...
+            $magentoRoot . '/docker/config.php.dist',
+        ];
+
+        foreach ($files as $filePath) {
+            if (!$this->file->isExists($filePath)) {
```
shiftedreality: Let's remove dist file from `ece-tools` and generate it every time once this command called

src/Util/CloudVariableEncoder.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\MagentoCloud\Util;
+
+/**
+ * Encodes and decodes data in Cloud format.
+ */
+class CloudVariableEncoder
```
shiftedreality: Let's move this to `src/Encoder/VariableEncoder.php`

shiftedreality (changes requested):

src/Util/CloudVariableEncoder.php
oshmyheliuk: Why? It's just helping class, I'm not sure that we need a new namespace for it.

src/Docker/Config/RelationshipUpdater.php
oshmyheliuk: Ok, good point

src/Util/CloudVariableEncoder.php
shiftedreality: The helper class is an anti-pattern, those utils should be in appropriate places respectively

src/Docker/Config/RelationshipUpdater.php
oshmyheliuk: Fixed.

src/Docker/Config/Relationship.php
oshmyheliuk: done

src/App/Container.php
```diff
@@ -146,6 +146,14 @@ function () {
             \Magento\MagentoCloud\View\RendererInterface::class,
             \Magento\MagentoCloud\View\TwigRenderer::class
         );
+        $this->container->singleton(
+            \Magento\MagentoCloud\CloudVariable\EncoderInterface::class,
+            \Magento\MagentoCloud\CloudVariable\Encoder::class
+        );
+        $this->container->singleton(
+            \Magento\MagentoCloud\CloudVariable\DecoderInterface::class,
```
shiftedreality: Decoder is excessive, it's ok to have only Encoder

src/CloudVariable/Encoder.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\MagentoCloud\CloudVariable;
```
shiftedreality: This namespace is not correct. You call namespace of purpose instead of calling implementation for purpose.
It should be:
```
Magetno\MagentoCloud\Encoder
-- PlatformEncoder
-- EncoderInterface
```

src/CloudVariable/Encoder.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Encodes data into Cloud format.
+ */
+class Encoder implements EncoderInterface
```
shiftedreality: You don't use this class. Add usages or remove it

src/CloudVariable/Encoder.php
oshmyheliuk: DecoderInterface that I added is resposible only for decoding cloud variable It shouldn't be in general `Encoder` space because it has specific purpose. It should be in some `CloudVariable` or just `Variable` space.

src/CloudVariable/Encoder.php
oshmyheliuk: Ok, will remove it

src/CloudVariable/Encoder.php
shiftedreality: your interface `DecoderInterface` does not belong to space `CloudVariable` -  it is a generic interface which supposes to be used for more decoding implementations.

oshmyheliuk: It's not a generic Decoder Interface it's a specific interface for decoding cloud variables. It has only one purpose, all implementation should be used only for decoding cloud variables.

shiftedreality: This does not make much sense, but in this case, rename namespace to `PlatformVariable`

oshmyheliuk: Done

shiftedreality (approved):

arhiopterecs (approved):

andriyShevtsov:
QA approved.

mveeramneni:
test message for github-slack

PR #475, opened 2019-05-01
MAGECLOUD-3475: Refactor Shell implementation

oshmyheliuk:
### Description

The current implementation of Shell does not expect to have non-zero output from the command. This affects few commands which can return different codes like setup:db:status.

In this PR changed the implementation of Shell command to return a result with output message and exit code.

https://magento2.atlassian.net/browse/MAGECLOUD-3475

### Manual testing scenarios

1. Deploy Magento to cloud.
2. Check that all shell commands are running as usual.

https://jira.corp.magento.com/browse/MAGECLOUD-117

### Contribution checklist


src/Shell/ResultInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MagentoCloud\Shell;
+
+/**
+ * Result for shell command.
+ */
+interface ResultInterface
```
shiftedreality: Why do we need this? I propose to return the instance of Process which has much more features.

src/Shell/Shell.php
```diff
@@ -53,70 +74,47 @@ public function __construct(LoggerInterface $logger, SystemList $systemList, San
...
-     * @param int $status
-     * @return array
+     * Logs command output
      *
-     * @throws ShellException
+     * @param Process $process
+     * @return void
      */
-    private function handleOutput(string $command, array $output, int $status): array
+    private function handleOutput(Process $process)
```
shiftedreality: This can be done as our Process which extends Symfony's Process and redeclares getOutput method

src/Shell/Result.php
```diff
@@ -0,0 +1,52 @@
...
+        return (int)$this->process->getExitCode();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getOutput(): array
+    {
+        try {
+            $output = explode(PHP_EOL, trim($this->process->getOutput(), PHP_EOL));
```
shiftedreality: The array is not the best return value here, let's keep it as a string by default

shiftedreality (changes requested):

src/Shell/ResultInterface.php
oshmyheliuk: Because it's not a good practice to make the high coupling between our code and third-party class.
If something will change in the third-party library it can break our code in many places. 
But when we use our interface all changes can be fixed in one class.

src/Shell/Result.php
oshmyheliuk: Some shell command as `which bash` can return multiple results and string will be like `/path1\n/path2\n/path3\n`.  It's not convenient to operate with such results.

src/Shell/Shell.php
oshmyheliuk: Extending external class is not the best idea it will make our code tightly coupled to external libraries.

src/Shell/Result.php
shiftedreality: Most commands will return only one string, having multiple is edge-case which must be handled accordingly

src/Shell/Shell.php
shiftedreality: We're coupling to external lib in one or another way

src/Shell/ResultInterface.php
shiftedreality: This library uses Semantic Versioning, we should rely on it

src/Shell/Result.php
oshmyheliuk: One more point: previously shell command was returning an array as a result and I just wanted to keep it consistent.

src/Shell/ResultInterface.php
oshmyheliuk: For example:  if magento 2.4 will require Symfony Console 4+ how Semantic Versioning will help us? some methods in Process class will be deprecated and we can use them at that moment in our code.

shiftedreality: We have a bunch of 3rd-party libs in use, once we'll see breaking changes - we can think about introducing some adapter-compatiblity

oshmyheliuk: At some point, it can be very hard to add an adapter or even impossible. Better to prevent such situations, especially if we want to add some extensibility in our code.

shiftedreality:
@oshmyheliuk please increase tests coverage

src/Process/DbDump/DbDump.php
```diff
@@ -110,9 +110,9 @@ public function execute()
                 $this->logger->info('Start creation DB dump...');
 
                 $command = $this->getCommand() . ' | gzip > ' . $dumpFile;
-                $errors = $this->shell->execute('bash -c "set -o pipefail; ' . $command . '"');
+                $process = $this->shell->execute('bash -c "set -o pipefail; ' . $command . '"');
 
-                if ($errors) {
+                if ($process->getExitCode() !== 0) {
```
shiftedreality: Let's add zero code constant to ShellInterface s `CODE_SUCCESS = 0;`

src/Shell/Process.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function execute()
+    {
+        try {
+            parent::mustRun();
+        } catch (ProcessFailedException $e) {
+            throw new \RuntimeException($e->getMessage(), $e->getProcess()->getExitCode());
```
shiftedreality: It must throw `ProcessException`

src/Shell/Shell.php
```diff
@@ -53,70 +64,48 @@ public function __construct(LoggerInterface $logger, SystemList $systemList, San
...
-
-        exec($fullCommand, $output, $status);
-
-        /**
-         * config:show will return non-zero exit code, if the value was not changed and remains default.
-         */
-        if ($status !== 0 && strpos($command, 'config:show') !== false) {
-            return [];
+            $process->execute();
+        } catch (\RuntimeException $e) {
```
shiftedreality: catch `ProcessException`

src/Shell/Shell.php
```diff
@@ -53,70 +64,48 @@ public function __construct(LoggerInterface $logger, SystemList $systemList, San
...
+                $command = array_merge([$command], $args);
+            }
 
-        if ($args) {
-            $command .= ' ' . implode(' ', $args);
-        }
+            $process = $this->processFactory->create([
+                'commandline' => $command,
+                'cwd' => $this->systemList->getMagentoRoot(),
+                'timeout' => 0
```
shiftedreality: it should be `null` according to docs

shiftedreality (changes requested):

src/Shell/Process.php
oshmyheliuk: Are you proposing to create a new Exception? Because we can't throw `\Magento\MagentoCloud\Process\ProcessException` here, as this Process is not related to `\Magento\MagentoCloud\Process`.

shiftedreality:
@oshmyheliuk please recreate this PR from main repo to trigger automated tests.

src/Shell/Shell.php
oshmyheliuk: fixed

src/Process/DbDump/DbDump.php
oshmyheliuk: added

PR #537, opened 2019-06-07
MAGECLOUD-3642: Time to First Byte Test

bbatsche:
### Description

Added process to do time to first byte testing for a user controlled list of URLs

[MAGECLOUD-3642](https://magento2.atlassian.net/browse/MAGECLOUD-3642)

### Manual testing scenarios

[MAGECLOUD-129](https://jira.corp.magento.com/browse/MAGECLOUD-129)

### Contribution checklist


src/Config/Stage/PostDeployInterface.php
```diff
@@ -13,4 +13,8 @@
 interface PostDeployInterface extends StageConfigInterface
 {
     const VAR_WARM_UP_PAGES = 'WARM_UP_PAGES';
+
+    const VAR_TTFB_TESTED_PAGES = 'TTFB_TESTED_PAGES';
+
+    const VAR_ENABLE_TTFB_TEST = 'ENABLE_TTFB_TEST';
```
oshmyheliuk: Looks like you forgot to remove this constant.

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
oshmyheliuk: Please remove empty line

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,64 @@
...
+        RequestFactory $requestFactory,
+        UrlManager $urlManager
+    ) {
+        $this->container = $container;
+        $this->clientFactory = $clientFactory;
+        $this->requestFactory = $requestFactory;
+        $this->urlManager = $urlManager;
+    }
+
+    public function create(array $urls, array $config, array $clientConfig = [], string $requestMethod = 'GET'): Pool
```
oshmyheliuk, line 67: Please add docs to all new function/classes

src/Config/Stage/PostDeployInterface.php
```diff
@@ -13,4 +13,6 @@
 interface PostDeployInterface extends StageConfigInterface
 {
     const VAR_WARM_UP_PAGES = 'WARM_UP_PAGES';
+
```
shiftedreality: remove newline

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,71 @@
...
+use Magento\MagentoCloud\App\ContainerInterface;
+use Magento\MagentoCloud\Util\UrlManager;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Create Guzzle request pools.
+ */
+class PoolFactory
+{
+    /** @var ContainerInterface */
```
shiftedreality: Don't one-line phpdocs

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,71 @@
...
+    /** @var ClientFactory */
+    private $clientFactory;
+
+    /** @var RequestFactory */
+    private $requestFactory;
+
+    /** @var UrlManager */
+    private $urlManager;
+
+    public function __construct(
```
shiftedreality, line 46: add phpdoc

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Generate Requests instances for a Pool.
+     */
+    public function yieldRequest(array $urls, string $requestMethod): \Iterator
+    {
+        foreach ($urls as $url) {
+            $url = $this->urlManager->expandUrl($url);
+
+            yield $this->requestFactory->create($requestMethod, $url);
```
shiftedreality: why do we need a generator here?

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Log stats from Guzzle request.
+     *
+     * @param TransferStats $stats
+     */
+    public function __invoke(TransferStats $stats)
+    {
+        $status = $stats->hasResponse() ? $stats->getResponse()->getStatusCode() : 'unknown';
+
+        if (300 < $status && $status < 400) {
```
shiftedreality, line 56: it seems more reasonable to pass only requests with `code == 200`

src/Process/PostDeploy/TimeToFirstByte.php
```diff
@@ -0,0 +1,98 @@
...
+            throw new ProcessException($exception->getMessage(), $exception->getCode(), $exception);
+        }
+    }
+
+    /**
+     * Returns list of URLs which should tested.
+     *
+     * @return array
+     */
+    public function getUrlsForTesting(): array
```
shiftedreality: This should be private method

shiftedreality: Name is too overloaded with context, please change to something simpler, like `getUrls`

src/Process/PostDeploy/TimeToFirstByte.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return array_filter(
+            $this->postDeploy->get(PostDeployInterface::VAR_TTFB_TESTED_PAGES),
+            function ($page) {
+                if (!$this->urlManager->isUrlValid($page)) {
+                    $this->logger->error(sprintf('Will not test %s, host is not a configured store domain', $page));
```
shiftedreality: It's warning, the application won't break

shiftedreality: Message can be simplified to `Url %s is not valid`

src/Process/PostDeploy/WarmUp.php
```diff
@@ -111,50 +95,25 @@ public function execute()
      *
      * @return array
      */
-    private function getUrlsForWarmUp(): array
+    public function getUrlsForWarmUp(): array
     {
```
shiftedreality: This must be private

src/Process/PostDeploy/WarmUp.php
```diff
@@ -44,63 +38,53 @@ class WarmUp implements ProcessInterface
...
+                    'code' => $exception->getResponse()->getStatusCode(),
+                ];
             }
 
-            \GuzzleHttp\Promise\unwrap($promises);
+            $this->logger->error('Warming up failed: ' . $urls[$index], $context);
+        };
+
+        try {
+            $pool = $this->poolFactory->create($urls, compact('fulfilled', 'rejected'));
```
shiftedreality, line 84: `compact` should be avoided

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            $time = $stats->getHandlerStats()[CURLINFO_STARTTRANSFER_TIME];
+        }
+
+        $this->logger->info(
+            sprintf('TTFB test result: %01.3fs', $time),
+            ['url' => (string) $stats->getEffectiveUri(), 'status' => $status]
+        );
+
+        $historicData = $this->file->isExists($this->fileList->getTtfbLog())
```
shiftedreality, line 73: Your historical file could possibly grow a lot. Reading it into memory and decoding would consume memory, I would consider writing it as plain text (a la log)

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,76 @@
...
+use Magento\MagentoCloud\Filesystem\Driver\File;
+use Magento\MagentoCloud\Filesystem\FileList;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Callback for Guzzle TransferStats to log their time to first byte value.
+ */
+class TransferStatsHandler
+{
+    /** @var File */
```
shiftedreality: Use long phpdoc notation everywhere

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
shiftedreality: no newline needed

shiftedreality (changes requested):

src/Http/PoolFactory.php
bbatsche: Why not? They're clean and easy to read.

src/Http/PoolFactory.php
bbatsche: What is there to document here? Parameter types are already type hinted.

src/Http/PoolFactory.php
bbatsche: I was working based off of the example code from Guzzle for creating concurrent requests, this should (for example) help running the warm-up process asynchronously. https://guzzle.readthedocs.io/en/stable/quickstart.html#concurrent-requests

src/Http/TransferStatsHandler.php
bbatsche: We want to log responses that are `2xx`, `4xx`, or `5xx`. Really the need here was to skip any redirects since they would wind up getting logged twice

src/Process/PostDeploy/TimeToFirstByte.php
bbatsche: Making this public doesn't harm anything and means we can unit test the method.

src/Process/PostDeploy/TimeToFirstByte.php
bbatsche: Good point, I'll update the log level in `WarmUp::getUrlsForWarmUp()` as well.

bbatsche: I guess? I feel like it's worth the extra text to tell the use _why_ the URL is not valid. But I don't have a strong opinion here

src/Process/PostDeploy/WarmUp.php
bbatsche: Why? Private methods cannot be unit tested.

src/Process/PostDeploy/WarmUp.php
bbatsche: Why? It's a safe and reliable method.

src/Http/TransferStatsHandler.php
bbatsche: That's a good concern. The goal was to dump into a JSON file so that it could be easily parsed by a third party utility. If we just dump into a flat text file there isn't much added utility above our existing log messages. Will think more about this

src/Process/PostDeploy/WarmUp.php
shiftedreality: You should not create API methods (which are protected and public) just for unit testing. Test original `execute` method or decouple class if it's to complex

src/Http/PoolFactory.php
shiftedreality: This violates coding standards, which ECE-Tools currently uses

src/Http/PoolFactory.php
shiftedreality: You should use PHP doc for all methods, even if it's type-hinted

src/Http/PoolFactory.php
shiftedreality: Because generators make code less readable and in your case without the visible advantage

src/Process/PostDeploy/TimeToFirstByte.php
shiftedreality: The same, you should not violate coding standards just for unit testing purposes

src/Process/PostDeploy/WarmUp.php
shiftedreality: It has broad visibility scope and makes code less readable

src/Http/PoolFactory.php
bbatsche: It's valid PHPDoc format and is already used in many places throughout both ECE Tools and other Magento packages (mostly in unit test classes). I can find nothing in https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html to suggest it violates our coding standards.

src/Process/PostDeploy/WarmUp.php
bbatsche: I'm not sure what you mean by "visibility scope"; `compact()` doesn't have any greater scope than other variables/functions. I also fundamentally disagree that
```php
['fulfilled' => $fulfilled, 'rejected' => $rejected]
```
is somehow more readable.

src/Http/PoolFactory.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function create(array $urls, array $config, array $clientConfig = [], string $requestMethod = 'GET'): Pool
+    {
+        $client = $this->clientFactory->create($clientConfig);
+
+        $requests = array_map(function ($url) use ($requestMethod) {
+            return $this->requestFactory->create($requestMethod, $this->urlManager->expandUrl($url));
+        }, $urls);
+
+        return $this->container->create(Pool::class, compact('client', 'requests', 'config'));
```
shiftedreality, line 75: Compact usage must be eliminated, it is not readable, not parsable by IDE's and with high probability will fail on PHPStan

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,81 @@
...
+     * @var FileList
+     */
+    private $fileList;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(File $file, FileList $fileList, LoggerInterface $logger)
```
shiftedreality, line 40: Please add PHPDoc

src/Http/TransferStatsHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $status = $stats->hasResponse() ? $stats->getResponse()->getStatusCode() : 'unknown';
+
+        if (300 < $status && $status < 400) {
+            $this->logger->debug('TTFB response was a redirect', ['url' => (string) $stats->getEffectiveUri()]);
+            return;
+        }
+
+        if (!array_key_exists(CURLINFO_STARTTRANSFER_TIME, $stats->getHandlerStats())) {
+            $this->logger->debug('cURL stats are missing from the request; using total transfer time');
```
shiftedreality, line 62: @hguthrie please review this message

src/Process/PostDeploy/WarmUp.php
shiftedreality: Answered on the previous `compact` usage

src/Test/Unit/Http/PoolFactoryTest.php
```diff
@@ -0,0 +1,109 @@
...
+     * @var UrlManager|MockObject
+     */
+    private $urlManagerMock;
+
+    /**
+     * @var PoolFactory
+     */
+    private $poolFactory;
+
+    protected function setUp()
```
shiftedreality, line 54: Add inheritdoc

src/Test/Unit/Http/TransferStatsHandlerTest.php
```diff
@@ -0,0 +1,152 @@
...
+     * @var LoggerInterface|MockObject
+     */
+    private $loggerMock;
+
+    /**
+     * @var TransferStatsHandler
+     */
+    private $handler;
+
+    public function setUp()
```
shiftedreality, line 45: Add inheritdoc

shiftedreality (dismissed):

src/Http/PoolFactory.php
bbatsche: It won't. PHPStan at the highest level with optional strict rules applied has no problem with `compact()`. I'm not sure what parsing your IDE needs to do with `compact()`, but if it's struggling with a function that has existed since PHP 4 and is 100% supported through the latest version, maybe you should consider getting a better IDE.

oshmyheliuk: I'm not a fan of  `compact` function, but I don't see any problems with using it.

dist/.magento.env.yaml
```diff
@@ -427,6 +427,17 @@
...
+# Magento Version: Magento 2.1.4 and later                                                                            #
+# Default value: - [] (an empty array)                                                                                #
+# Stages: post-deploy                                                                                                 #
+# Example:                                                                                                            #
+#          stage:                                                                                                     #
+#            post-deploy:                                                                                             #
+#              TTFB_TESTED_PAGES:                                                                                     #
+#                - "index.php"                                                                                        #
+#                - "index.php/customer/account/create"                                                                #
```
oshmyheliuk, line 439: It'll be good to add some page examples with a host name.  Or add this to the description.

src/Test/Unit/Process/PostDeploy/TimeToFirstByteTest.php
```diff
@@ -0,0 +1,113 @@
...
+            ->with(
+                $urls,
+                ['options' => [RequestOptions::ON_STATS => $this->statHandlerMock], 'concurrency' => 1]
+            )->willReturn($this->poolMock);
+        $this->promiseMock->expects($this->once())
+            ->method('wait');
+
+        $this->process->execute();
+    }
+}
```
billygilbert, line 140: Should test for exception here.

billygilbert (dismissed):

oshmyheliuk (dismissed):

billygilbert (dismissed):

billygilbert (approved):

oshmyheliuk (approved):

mveeramneni:
QA Approved

PR #546, opened 2019-06-19
Reduce log level of shell output

michaelcasey316:
Reduce log level of shell output

### Description
This pull request reduces the log level of shell output from INFO to DEBUG. The corresponding unit tests are also modified to match.

1. magento/ece-tools/issues/467

### Manual testing scenarios
1. Deploy to a remote integration environment.
2. See that the deploy log is less noisy.

### Contribution checklist
 - [x] All automated tests passed successfully (all builds on Travis CI are green)


src/Test/Unit/Shell/ShellTest.php
```diff
@@ -188,10 +187,10 @@ public function testExecuteException()
             ->method('sanitize')
             ->with($this->stringContains('ls -al --password="123"'))
             ->willReturn('Command ls -al --password="***" failed');
-        $this->loggerMock->expects($this->once())
+        $this->loggerMock->expects($this->never())
             ->method('info')
             ->with($command);
```
oshmyheliuk: Please move `->with($command)` from `info` method mock to `debug` method  mock.
And you can remove `info`  method mock at all as it never uses in the `Shell` class.

oshmyheliuk:
Several functional tests failed which depend on shell command log message. 
We can reduce log level to `debug` for such tests:
```
stage:                                                                                                    
    global:                                                                                                 
         MIN_LOGGING_LEVEL: debug          
```

src/Test/Unit/Shell/ShellTest.php
michaelcasey316: Great point, thanks for catching this.

src/Test/Functional/Acceptance/PatchApplierCest.php
```diff
@@ -33,7 +33,6 @@ public function testApplyingPatch(\CliTester $I)
         $I->assertContains('## Additional Info', $targetFile);
         $log = $I->grabFileContent('/var/log/cloud.log', Docker::BUILD_CONTAINER);
         $I->assertContains('INFO: Applying patch /var/www/magento/m2-hotfixes/patch.patch', $log);
-        $I->assertContains('INFO: git apply /var/www/magento/m2-hotfixes/patch.patch', $log);
```
oshmyheliuk, line 36: Maybe is better not to remove this line but change `INFO` -> `DEBUG` and add 
```
$I->uploadToContainer('files/debug_logging/.magento.env.yaml', '/.magento.env.yaml', Docker::BUILD_CONTAINER);
```
as for `AdminCredentialCest`

michaelcasey316: I thought about this, but thought it wasn't as important because of the line directly above it which essentially tests the same thing:
` $I->assertContains('INFO: Applying patch /var/www/magento/m2-hotfixes/patch.patch', $log);`

Do you think it's important to test the "git apply" output in addition to the "applying patch" output, even if they both have the same filename?

oshmyheliuk: I'm not sure if it's important, but if Author of the test added this check he thought it was important.

michaelcasey316: I'll add it back then. Thanks for the quick reply.

oshmyheliuk (approved):

NadiyaS (approved):

BaDos (approved):

mveeramneni:
QA Approved

PR #561, opened 2019-07-09
MAGECLOUD-3816: Eliminate Docker config.env file

shiftedreality:

1. https://magento2.atlassian.net/browse/MAGECLOUD-3816

### Manual testing scenarios

1. https://jira.corp.magento.com/browse/MAGECLOUD-66
2. https://jira.corp.magento.com/browse/MAGECLOUD-116

### Contribution checklist


src/Command/Docker/ConfigConvert.php
```diff
@@ -1,108 +0,0 @@
...
-use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Output\OutputInterface;
-
-/**
- * Converts raw array configuration to .env files.
- */
-class ConfigConvert extends Command
-{
-    const NAME = 'docker:config:convert';
```
NadiyaS, line 22: you remove this command at all?
if so, it is BIC. You need to leave this command and result of it should be output with explanation what to use instead

src/Command/Docker/GenerateDist.php
```diff
@@ -0,0 +1,60 @@
...
+     * @inheritdoc
+     */
+    protected function configure()
+    {
+        $this->setName(self::NAME)
+            ->setDescription('Generates Docker .dist files');
+    }
+
+    /**
+     * @inheritDoc
```
NadiyaS: cover @inheritdoc with {}

src/Docker/Config/Environment/Reader.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Magento\MagentoCloud\Docker\Config\Environment;
+
+use Magento\MagentoCloud\Docker\ConfigurationMismatchException;
+use Magento\MagentoCloud\Filesystem\DirectoryList;
+use Magento\MagentoCloud\Filesystem\Driver\File;
+use Magento\MagentoCloud\Filesystem\SystemList;
+
+/**
+ * Reader for .env files.
```
NadiyaS: Seems like description is not correct

src/Command/Docker/GenerateDist.php
shiftedreality: Fixed

NadiyaS (changes requested):
We need to define the BC rules.
Cause what I see now is:
 - removing CLI command
 - removing exception type

what is BIC as for me, especially the first one

src/Command/Docker/ConfigConvert.php
shiftedreality: It is a manual run command, which can be excluded from BC list

shiftedreality:
Exceptions in ECE-Tools are not part of BC rules

src/Command/Docker/GenerateDist.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function __construct(Generator $distGenerator)
+    {
+        $this->distGenerator = $distGenerator;
+
+        parent::__construct();
+    }
+
+    /**
+     * {@inheritDoc}
```
BaDos: You should use `{}` only if you expand the description.

For example:
```
/**
 * @inheritDoc
 */
```

```
/**
 * {@inheritDoc}
 * @return bool
 */
```

BaDos: Please, fix all places with `@inheritDoc`

BaDos: Also, I see that usually, we use `@inheritdoc`, you write `@inheritDoc`
I think we should use one style.

BaDos (changes requested):
Please, fix all @inheritdoc

src/Command/Docker/GenerateDist.php
shiftedreality: There is no clear explanation of `@inheritDoc`, until that, I don't see a big reason to point attention on it.

src/Command/Docker/ConfigConvert.php
NadiyaS: hm. Manual command may be used in scripts by others to automate deployment on cloud docker, for example. And even if use it only manually, it is not good strategy to surprise people with absence of some commands after new release.
I agree with removing Extension in current moment as we do not provide BC for code itself, but removing/renaming commands or change their behavior is BIC IMHO.

shiftedreality: Added name alias

BaDos (dismissed):

NadiyaS (approved):

BaDos (approved):

andriyShevtsov:
QA approved

PR #567, opened 2019-07-18
MAGECLOUD-3578  Add PHPStan to Static Tests set

oshmyheliuk:
### Description
Added PHPStan to Static Tests set

https://magento2.atlassian.net/browse/MAGECLOUD-3578

### Contribution checklist


src/Config/Schema.php
```diff
@@ -79,7 +79,7 @@ public function getSchema()
                 self::SCHEMA_VALUE_VALIDATION => function (string $key, $value) {
                     if (!in_array($value, range(0, 9))) {
                         return sprintf(
-                            'The SCD_COMPRESSION_LEVEL variable contains an invalid value of type string. ' .
+                            'The %s variable contains an invalid value #%d. ' .
```
bbatsche: I'm not sure the `#` actually makes sense in this context, I would leave it out

src/Docker/Compose/DeveloperCompose.php
```diff
@@ -30,7 +30,8 @@ public function build(Repository $config): array
     {
         $compose = parent::build($config);
         $compose['volumes'] = [
-            'magento-sync' => self::SYNC_ENGINE_DOCKER_SYNC === $config[self::SYNC_ENGINE] ? ['external' => true] : []
+            'magento-sync' => self::SYNC_ENGINE_DOCKER_SYNC === $config->get(self::SYNC_ENGINE) ?
```
bbatsche, line 33: I personally think this format's a lot easier to read across three lines:
```php
'magento-sync' => self::SYNC_ENGINE_DOCKER_SYNC === $config[self::SYNC_ENGINE]
    ? ['external' => true]
    : []
```

but that's a matter of preference

src/Config/Schema.php
bbatsche: The `#` is really unnecessary here, I would leave it off

tests/static/phpstan.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+	excludes_analyse:
```
bbatsche: It may be easier/more consistent to add:
```yaml
  level: 5
  paths:
    - src
```
This way there's less to worry about passing on the command line by default

tests/static/phpstan.neon
```diff
@@ -0,0 +1,6 @@
...
+		- %currentWorkingDirectory%/src/Test/*
+	ignoreErrors:
+		- message: "#Strict comparison using === between bool and null will always evaluate to false.#"
```
bbatsche: This seems like something we should actually fix, maybe an issue for our backlog

composer.json
```diff
@@ -38,7 +38,8 @@
     "phpunit/phpunit": "^6.2",
     "squizlabs/php_codesniffer": "^3.0",
     "codeception/codeception": "^2.5.3",
-    "consolidation/robo": "^1.2"
+    "consolidation/robo": "^1.2",
+    "phpstan/phpstan": "@stable"
```
bbatsche, line 42: Did you look at adding [`phpstan/phpstan-phpunit`](https://github.com/phpstan/phpstan-phpunit)? Does it make static analysis of our unit tests any more feasible?

bbatsche (changes requested):

composer.json
oshmyheliuk: No, I didn't try, unit tests was excluded from phpstan checking as they contain too many issues. We can add a task  about this to our backlog.

tests/static/phpstan.neon
oshmyheliuk: This message is ignoring only in one class `File.php` as we suppress warning `@` of native php functions and PHPstan doesn't expect that function can return a null value.
I don't know how we can easily fix this.

src/Config/Schema.php
oshmyheliuk: agree, removed

tests/static/phpstan.neon
oshmyheliuk: Make sense, updated. Thank you.

oshmyheliuk: This only works for the newer versions of PHPstan and doesn't work for versions compatible with PHP 7.0. So I didn't add level and paths to `phpstan7.0.neon` file.

arhiopterecs (approved):

bbatsche (approved):

shiftedreality:
Proposed changes do not affect business functionality. No QA needed.

PR #569, opened 2019-07-22
MAGECLOUD-3940: Use more OS agnostic form to find number of processors

keithbentrup:
nproc is the better, more common way to find the number of processors (CPUs). It exists for linux and OSX. This change helps developers run "ece-tools deploy" on OSX faster while preparing their code for cloud.

1. on OSX, ece-tools does not properly detect # of CPUs

### Manual testing scenarios
1. run `nproc` on cloud; 
2. see that the result matches `grep -c processor /proc/cpuinfo`

* https://jira.corp.magento.com/browse/MAGECLOUD-44
* https://jira.corp.magento.com/browse/MAGECLOUD-53

### Contribution checklist
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/src/Util/Cpu.php
+++ b/src/Util/Cpu.php
@@ -45,7 +45,7 @@ class Cpu
     public function getThreadsCount(): int
     {
         try {
-            $result = $this->shell->execute('grep -c processor /proc/cpuinfo');
+            $result = $this->shell->execute('nproc');
             $threadCount = (int)$result->getOutput() ?? 1;
         } catch (ShellException $e) {
             $this->logger->error('Can\'t get system processor count: ' . $e->getMessage());
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,8 @@ matrix:
       env: TEST_SUITE=functional
 
 install:
-  - if [ $TRAVIS_SECURE_ENV_VARS == "false" ]; then composer remove magento/magento-cloud-components --no-update; fi;
-  - composer config http-basic.repo.magento.com ${REPO_USERNAME} ${REPO_PASSWORD}
+  - if [ $TRAVIS_SECURE_ENV_VARS != "true" ]; then composer remove magento/magento-cloud-components --no-update; fi;
+  - if [ $TRAVIS_SECURE_ENV_VARS == "true" ]; then composer config http-basic.repo.magento.com ${REPO_USERNAME} ${REPO_PASSWORD}; fi;
   - composer update -n --no-suggest
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
 
 install:
   - if [ $TRAVIS_SECURE_ENV_VARS != "true" ]; then composer remove magento/magento-cloud-components --no-update; fi;
-  - if [ $TRAVIS_SECURE_ENV_VARS != "true" ]; then composer config --unset repositories.repo.magento.com fi;
+  - if [ $TRAVIS_SECURE_ENV_VARS != "true" ]; then composer config --unset repositories.repo.magento.com; fi;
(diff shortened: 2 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/ece-tools?pullRequest=569) <br/>All committers have signed the CLA.

src/Util/Cpu.php
```diff
@@ -45,7 +45,7 @@ public function __construct(ShellInterface $shell, LoggerInterface $logger)
     public function getThreadsCount(): int
     {
         try {
-            $result = $this->shell->execute('grep -c processor /proc/cpuinfo');
+            $result = $this->shell->execute('nproc');
```
oshmyheliuk, line 48: Please update Unit test for this class

andriyShevtsov:
Created  internal ticket https://magento2.atlassian.net/browse/MAGECLOUD-3940

src/Util/Cpu.php
keithbentrup: @oshmyheliuk I'm not sure what you mean - which unit test? Can you link to what you mean?

oshmyheliuk: After your changes this Unit test `\Magento\MagentoCloud\Test\Unit\Util\CpuTest` will fail. 
Please update it here `src/Test/Unit/Util/CpuTest.php:54`.
```
    public function testGetTreadsCount()
    {
        $processMock = $this->getMockForAbstractClass(ProcessInterface::class);
        $processMock->expects($this->once())
            ->method('getOutput')
            ->willReturn('8');
        $this->shellMock->expects($this->once())
            ->method('execute')
            ->with('grep -c processor /proc/cpuinfo')
            ->willReturn($processMock);

        $this->assertEquals(8, $this->cpu->getThreadsCount());
    }
```
Thank you.

keithbentrup: @oshmyheliuk Hmm, maybe I'm misunderstanding what this unit test is supposed to accomplish, but I think that the unit test itself needs some code review. It only tests that the string matches. You could literally put anything in there, and it would pass as long as the strings match. For example:
image
So what's the point of the test? Shouldn't the test actually validate that you have the right number of cpus? (which is exactly what `nproc` returns and is intended for (and probably should have been used in the first place had the original author known about it or if the author did know about `nproc`, the author should have commented why they were choosing not to use it).

Anyway, if you want me to fix testGetTreadsCount, too, my solution would be to get rid of the test altogether because it provides no value and "unit testing" a single word system command with no parameters doesn't make much sense and just adds more lines of code that need to be maintained or can introduce errors (or have to be justified). In fact, I would get rid of the entire Cpu class if its only purpose is to return what `nproc` tells you in a far less error prone way than grepping /proc/cpuinfo with a regular expression. All in all with Cpu.php and CPUTest.php, there's about ~100 loc to do what `nproc` does in a single word more reliably.

Also, there's a typo in `testGetTreadsCount`.

shiftedreality: Hello @keithbentrup. We use the unit tests as the second factor for code review, to ensure that code works as expected by the author. This gives us a huge advantage with merging public pull requests when there is a lack of communication.  Also, unit tests help to reduce the code complexity significantly. Please fix the failed test or let us know so we can fix it by ourselves in your pull request.

Thanks.

keithbentrup: Hi Oleh, I'm not sure that you read my response carefully. I understand the point of unit tests. That unit test does NOT "ensure that code works as expected". I literally replaced the `grep -c processor /proc/cpuinfo` with `this is not even a cmd`, and it "passed" the unit test. See my included screenshot. This is a very questionable test written for questionable code where the author was unaware of the proposed alternative. 

As I said before, neither this class nor the test should even exist, but I'm not particularly committed to having it removed at this point. I am disappointed that this is not immediately obvious to you or @oshmyheliuk. 

I'll reach out to you on slack to see if you have any questions.

oshmyheliuk (approved):

shiftedreality (approved):

mveeramneni:
QA Approved

Issue #534, opened 2019-06-07
[Cloud Docker] Set Varnish as default cache


YPyltiai:
### Preconditions

1. 
2. 

### Steps to reproduce

1. 
2. 
3. 

### Expected result

1. [Screenshots, logs or description]

### Actual result

1. [Screenshots, logs or description]

PR #573, opened 2019-07-24
MAGECLOUD-3598 [Cloud Docker]: Set Varnish as default cache

michaelcasey316:
### Description
Automatically select Varnish as the cache provider when starting Magento from cloud docker.

1. https://github.com/magento/ece-tools/issues/534

### Manual testing scenarios

1. Start Magento cloud docker as described on https://devdocs.magento.com/guides/v2.3/cloud/docker/docker-config.html
2. Verify the current full page caching application by running the following command:
`docker-compose run deploy magento-command config:show system/full_page_cache/caching_application`

### Contribution checklist

### DIFF
--- a/src/Docker/Compose/ProductionCompose.php
+++ b/src/Docker/Compose/ProductionCompose.php
@@ -376,6 +376,8 @@ class ProductionCompose implements ComposeInterface
             'PHP_IDE_CONFIG' => 'serverName=magento_cloud_docker',
             # Docker host for developer environments, can be different for your OS
             'XDEBUG_CONFIG' => 'remote_host=host.docker.internal',
+            # Tell Magento to automatically use the Varnish cache
+            'CONFIG__DEFAULT__SYSTEM__FULL_PAGE_CACHE__CACHING_APPLICATION' => '2',
         ], $this->reader->read());
     }
 
--- a/src/Docker/Compose/ProductionCompose.php
+++ b/src/Docker/Compose/ProductionCompose.php
@@ -32,6 +32,8 @@ class ProductionCompose implements ComposeInterface
 
     const CRON_ENABLED = true;
 
+    const VARNISH_CACHE_APPLICATION_ID = '2';
+
     /**
      * @var ServiceFactory
      */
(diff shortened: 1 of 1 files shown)


src/Docker/Compose/ProductionCompose.php
```diff
@@ -376,6 +378,8 @@ protected function getVariables(): array
             'PHP_IDE_CONFIG' => 'serverName=magento_cloud_docker',
             # Docker host for developer environments, can be different for your OS
             'XDEBUG_CONFIG' => 'remote_host=host.docker.internal',
+            # Tell Magento to automatically use the Varnish cache
+            'CONFIG__DEFAULT__SYSTEM__FULL_PAGE_CACHE__CACHING_APPLICATION' => self::VARNISH_CACHE_APPLICATION_ID,
```
shiftedreality: This may not work if Varnish service is removed from `docker-compose`. Please check this case

shiftedreality (changes requested):

src/Docker/Compose/ProductionCompose.php
michaelcasey316: I considered this possibility, but I don't think it's a big issue. We already assume other services, like the DB, are part of docker-compose. Also, the Varnish service is always added by the Docker config generator. As it stands, there's no way for it to be removed unless someone manually edits docker-compose.yml. 

Would it be better if the environment variable is only added if the Varnish config is part of `$services`?

shiftedreality: The Varnish service is not required to run Magento, so technically it's possible now and is a valid case

shiftedreality: On the other hand, this may be too complicated to implement, so do not spend too much time on investigation

YPyltiai: Well, a lot of people edit config manually, so I would account for that if possible. Let's see if we can set it up only when we have Varnish in place

michaelcasey316: @YPyltiai I'll look for alternative solutions to this problem, then.

PR #588, opened 2019-08-15
MAGECLOUD-3873 - Static Content Copy Deprecation Warning

bbatsche:
### Description

Add a warning if the user sets deprecated variable `STATIC_CONTENT_SYMLINK` to `false`.

[MAGECLOUD-3873](https://magento2.atlassian.net/browse/MAGECLOUD-3873)

### Manual testing scenarios

1. Add the following to `.magento.env.yaml`:

    ```yaml
    stage:
      deploy:
        STATIC_CONTENT_SYMLINK: false
    ```

1. Look for deprecation warning in the cloud.log file.

### Contribution checklist


src/Config/Validator/Deploy/DeprecatedVariables.php
```diff
@@ -69,7 +85,7 @@ public function validate(): Validator\ResultInterface
         }
 
         if ($this->environment->getEnv(DeployInterface::VAR_STATIC_CONTENT_THREADS)
-            || isset($variables[DeployInterface::VAR_STATIC_CONTENT_THREADS])
```
oshmyheliuk: It's not correct to change `$variables` to `$config` here, as `$config` is a merged config and VAR_STATIC_CONTENT_THREADS already removed from merged config
@see src/Config/Stage/Deploy/EnvironmentConfig.php:104 .
The same for VAR_DO_DEPLOY_STATIC_CONTENT variable
@see src/Config/Stage/Deploy/EnvironmentConfig.php:86

src/Config/Validator/Deploy/DeprecatedVariables.php
```diff
@@ -79,7 +95,7 @@ public function validate(): Validator\ResultInterface
         }
 
         if ($this->environment->getEnv(DeployInterface::VAR_DO_DEPLOY_STATIC_CONTENT)
-            || isset($variables[DeployInterface::VAR_DO_DEPLOY_STATIC_CONTENT])
+            || isset($config[DeployInterface::VAR_DO_DEPLOY_STATIC_CONTENT])
```
oshmyheliuk: See comment above.

oshmyheliuk (dismissed):

src/Config/Validator/Deploy/DeprecatedVariables.php
```diff
@@ -60,7 +73,10 @@ public function validate(): Validator\ResultInterface
             );
         }
 
-        if (isset($variables[DeployInterface::VAR_SCD_EXCLUDE_THEMES])) {
+        // default value for SCD_EXCLUDE_THEMES is an empty string
```
shiftedreality, line 76: `if (!empty())` covers both cases

shiftedreality (dismissed):

src/Config/Validator/Deploy/DeprecatedVariables.php
```diff
@@ -88,6 +104,15 @@ public function validate(): Validator\ResultInterface
             );
         }
 
+        if (isset($config[DeployInterface::VAR_STATIC_CONTENT_SYMLINK])
+            && $config[DeployInterface::VAR_STATIC_CONTENT_SYMLINK] === false
```
NadiyaS, line 106: why do you check only false value? probably this notification is better to add when variable exists with any value

src/Config/Validator/Deploy/DeprecatedVariables.php
```diff
@@ -60,7 +73,10 @@ public function validate(): Validator\ResultInterface
             );
         }
 
-        if (isset($variables[DeployInterface::VAR_SCD_EXCLUDE_THEMES])) {
+        // default value for SCD_EXCLUDE_THEMES is an empty string
+        if (isset($config[DeployInterface::VAR_SCD_EXCLUDE_THEMES])
+            && $config[DeployInterface::VAR_SCD_EXCLUDE_THEMES] !== ''
```
NadiyaS: If you want to add notification only when variable has some value then you also need to include empty array.
But I would notify customer in any case if this variable exists with any value. The same as for VAR_STATIC_CONTENT_SYMLINK

src/Test/Unit/Config/Validator/Deploy/DeprecatedVariablesTest.php
```diff
@@ -82,45 +94,59 @@ public function executeDataProvider(): array
...
             ],
             [
+                [],
                 [],
                 [DeployInterface::VAR_DO_DEPLOY_STATIC_CONTENT => 1],
                 Error::class,
             ],
             [
+                [],
                 [DeployInterface::VAR_SCD_EXCLUDE_THEMES => 'theme'],
```
NadiyaS, line 140: add cases with empty value for VAR_SCD_EXCLUDE_THEMES and true value for VAR_STATIC_CONTENT_SYMLINK

NadiyaS (changes requested):
Add all test cases, if you want to leave the validation of value, not only existence of variables.

src/Config/Validator/Deploy/DeprecatedVariables.php
bbatsche: `$config` comes from `MergedConfig` which includes the variables' default values (`true`), so we are in effect testing if the variable has been changed from its default state.

src/Config/Validator/Deploy/DeprecatedVariables.php
bbatsche: Again though, `MergedConfig` includes this variable's default value if none is set by the user. Testing if it's not `''` allows us to check that the variable has been modified.

src/Config/Validator/Deploy/DeprecatedVariables.php
NadiyaS: hm. Then we should remove default value and in the src/Filesystem/RecoverableDirectoryList.php just check the false value, otherwise behave like we would have true

NadiyaS: oh, I see that we will have an exception in this case. Not good

NadiyaS (approved):

shiftedreality (approved):

andriyShevtsov:
QA approved

PR #601, opened 2019-09-16
MAGECLOUD-3392: Reduce q-ty of Reports Created in /app/*/var/report

arhiopterecs:
### Description
https://magento2.atlassian.net/browse/MAGECLOUD-3392
Creates an association between the message in the exception log and the report; reduces the number of reports, keeping only unique reports; avoids overflowing the directory <magento_root>var/reports with millions of files with help additional subdirectories

1. magento/ece-tools#MAGECLOUD-3392: Reduce q-ty of Reports Created in /app/*/var/report

### Manual testing scenarios
1. https://jira.corp.magento.com/browse/MC-20162
2. https://jira.corp.magento.com/browse/MC-19586
3. https://jira.corp.magento.com/browse/MAGECLOUD-138

### Contribution checklist


patches.json
```diff
@@ -180,6 +180,18 @@
     "Re-work consumers to terminate as soon as there is nothing left to process": {
       "2.2.0 - 2.3.1": "MAGECLOUD-4071__terminate_consumers_if_the_queue_is_empty__2.2.0.patch",
       "2.3.2 - 2.3.3": "MAGECLOUD-4071__terminate_consumers_if_the_queue_is_empty__2.3.2.patch"
+    },
+    "Reduce q-ty of error report files": {
+      "2.1.4 - 2.1.7": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.4.patch",
+      "2.1.8 - 2.1.16": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.8.patch",
+      "2.1.17 - 2.1.18": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.17.patch",
```
NadiyaS: will Magento have this fix in 2.1.19?

oshmyheliuk: I changed patches constraints to ~2.1.17 and ~2.2.8 as backport wasn't merged into 2.2.

oshmyheliuk (dismissed):

YPyltiai:
Do not merge please. Auto-configuration approach is not clear - need some clarification.

patches.json
NadiyaS: is it planned to be merged? if yes, then we will need to change this right after that Magento release

YPyltiai: @NadiyaS not until after the configuration flow is fixed. Just adding patches doesn't solve the issue itself - so there is additional work required on this task.

dist/.magento.env.yaml
```diff
@@ -635,3 +635,25 @@
 #            variables:                                                                                               #
 #              ENV_ENVIRONMENT: "MAGENTO_CLOUD_ENVIRONMENT"                                                           #
 #######################################################################################################################
+# ERROR_REPORT_DIR_NESTING_LEVEL - The number of subdirectories that will be created to save the report.              #
+#                                                                                                                     #
+# If we have the report name as hash sha256('') = 44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7    #
```
NadiyaS: I would add link to the documentation about this configuration or if it was not released yen, then copy text which should be there

src/Config/Schema.php
```diff
@@ -284,6 +285,15 @@ public function getSchema()
...
                 ],
             ],
+            BuildInterface::VAR_ERROR_REPORT_DIR_NESTING_LEVEL => [
+                self::SCHEMA_TYPE => ['integer'],
+                self::SCHEMA_VALUE_VALIDATION => range(0, 32),
+                self::SCHEMA_STAGE => [
+                    StageConfigInterface::STAGE_GLOBAL,
+                    StageConfigInterface::STAGE_BUILD,
+                ],
+                self::SCHEMA_DEFAULT_VALUE => [StageConfigInterface::STAGE_BUILD => 1]
```
NadiyaS, line 295: @YPyltiai I am not sure that it is correct to change the report creation for everyone. It looks like BIC

YPyltiai: @NadiyaS it will go into minor ECE-Tools version (2002.1.0), so should be OK given how much issues it solves in general.

src/Config/Stage/BuildInterface.php
```diff
@@ -14,4 +14,8 @@
  */
 interface BuildInterface extends StageConfigInterface
 {
+    /**
+     * The number of subdirectories that will be created to save the report
```
NadiyaS: it is not a number, it is a variable name to provide number ;)

src/Process/Build/SetErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,100 @@
...
+
+use Magento\MagentoCloud\Config\Stage\BuildInterface;
+use Magento\MagentoCloud\Filesystem\ConfigFileList;
+use Magento\MagentoCloud\Filesystem\Driver\File;
+use Magento\MagentoCloud\Process\ProcessException;
+use Magento\MagentoCloud\Process\ProcessInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @inheritdoc
```
NadiyaS: Please increase the description

src/Process/Build/SetErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function execute()
+    {
+        $this->logger->info('Configuring directory nesting for saving error reports');
+        try {
+            $errorReportConfigFile = $this->configFileList->getErrorReportConfig();
+            if ($this->file->isExists($errorReportConfigFile)) {
```
NadiyaS: Is it possible that customers have this file, but without this variable?

NadiyaS: I see that `pub/errors/local.xml.sample` is not new file and it is quite possible that users already have it without `.sample`
So, need to parse this xml and check variable on existence

NadiyaS: So, there are several cases when we need to set, override or left this value as it is.
1. Customer does not have file:
 1.1 No ERROR_REPORT_DIR_NESTING_LEVEL variable in `env.yaml` - we create file and set default value 1
 1.2 Variable exists in `env.yaml` - we create file and set custom value from ERROR_REPORT_DIR_NESTING_LEVEL var
2. Customer has file and does not have this variable
 2.1 No ERROR_REPORT_DIR_NESTING_LEVEL variable in `env.yaml` - we set default value 1
 2.2 Variable exists in `env.yaml` - we set custom value from ERROR_REPORT_DIR_NESTING_LEVEL var
3. Customer has file and field `dir_nesting_level` exists in it:
 3.1 No ERROR_REPORT_DIR_NESTING_LEVEL variable in `env.yaml` - we **DO NOT** set default value. Value should be left as it is.
 3.2 Variable exists in `env.yaml` - we override custom value from ERROR_REPORT_DIR_NESTING_LEVEL var

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,75 @@
...
+            if ($envReportDirNestingLevel) {
+                $this->logger->notice(
+                    sprintf(
+                        '%s environment variable detected. %s: %s. '
+                        .'Value of the property \'config.report.dir_nesting_level\' from \'%s\' file be ignored',
+                        self::ENV_MAGE_ERROR_REPORT_DIR_NESTING_LEVEL,
+                        self::ENV_MAGE_ERROR_REPORT_DIR_NESTING_LEVEL,
+                        $envReportDirNestingLevel,
+                        $this->configFileList->getErrorReportConfig()
+                    )
```
NadiyaS: First, I would write this message only if customer define value for this variable in `env.yaml` file.
Also, this notice message duplicate the mane of variable, please rephrase it. E.g.
`Environment variable MAGE_ERROR_REPORT_DIR_NESTING_LEVEL exists with value {value}. Value of ERROR_REPORT_DIR_NESTING_LEVEL  defined in .magento.app.yaml will be ignored.`

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\MagentoCloud\Process\Deploy;
```
NadiyaS: why it is not in Magento\MagentoCloud\Config\Validator\Deploy namespace?

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,75 @@
...
+use Magento\MagentoCloud\Config\Environment;
+use Magento\MagentoCloud\Filesystem\ConfigFileList;
+use Magento\MagentoCloud\Process\ProcessException;
+use Magento\MagentoCloud\Process\ProcessInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @inheritdoc
+ */
+class CheckEnvVarMageErrorReportDirNestingLevel implements ProcessInterface
```
NadiyaS: I would rename this class to more readable one e.g. just ReportDirNesting or ReportDirNestingLevel.
No need to add `Check` or `Validator` in class name as it should be in `Magento\MagentoCloud\Config\Validator\Deploy` namespace

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,75 @@
...
+        Environment $environment,
+        ConfigFileList $configFileList
+    ) {
+        $this->logger = $logger;
+        $this->environment = $environment;
+        $this->configFileList = $configFileList;
+    }
+
+    /**
+     * @inheritdoc
```
NadiyaS: add Description. Or at least add description to class

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * @var Environment
+     */
+    private $environment;
```
NadiyaS: What is Environment? Please add more clear description

scenario/deploy.xml
```diff
@@ -6,6 +6,7 @@
  */
 -->
 <scenario xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../config/scenario.xsd">
+    <step name="check-evn-variable-mage-error-report-dir-nesting-level" type="Magento\MagentoCloud\Process\Deploy\CheckEnvVarMageErrorReportDirNestingLevel"/>
```
oshmyheliuk: For me, looks like the class name contains too many words. I would change to something like this

```
<step name="check-env-nesting-level" type="Magento\MagentoCloud\Process\Deploy\CheckEnvNestingLevel"/>
```

NadiyaS: move this to `validate-config` step 
and rename this step to the shortest version, please (e.g. report-dir-level, report-nesting-level, ... )

src/Config/Schema.php
NadiyaS: @YPyltiai if customer updates `metapackage` (not `ece-tools`) as we recommend, then ece-tools also be updated to the latest minor version

YPyltiai: @NadiyaS that should be OK in this case anyway - problem with this task is that site can go down completely if this is not fixed. Given you have a variable to disable this behavior (new one), we should be OK to proceed, since otherwise risks are higher and sites will continue going down because of incorrect report generation logic.

I think @kandy was also OK with this approach, since we don't have other options to fix it really.

Developers anyway look at what they upgrade, so minor ECE-Tools version bump won't go unnoticed. Same applies to any other minor version - metapackage updates will update other packages in exactly the same way

dist/.magento.env.yaml
arhiopterecs: @NadiyaS  The property ERROR_REPORT_DIR_NESTING_LEVEL is described in the PR https://github.com/magento/devdocs/pull/5485 (https://magento2.atlassian.net/browse/MAGECLOUD-4327).
A page with a detailed description was created "Directories Nesting Level Changes for Error Reporting" and a description about the property ERROR_REPORT_DIR_NESTING_LEVEL was added to the page "Building Variables"

src/Config/Stage/BuildInterface.php
arhiopterecs: Changed

src/Process/Build/SetErrorReportDirNestingLevel.php
arhiopterecs: Done

src/Process/Build/SetErrorReportDirNestingLevel.php
arhiopterecs: @NadiyaS 
Following a conversation with @YPyltiai, the following cases were agreed:

1. Customer does not have file:
1.1 No ERROR_REPORT_DIR_NESTING_LEVEL variable in env.yaml - we create a file and set default value 1
1.2 Variable exists in env.yaml - we create a file and set a custom value from ERROR_REPORT_DIR_NESTING_LEVEL var

2. If the client has a file, then we do not modify it.
At the deployment stage, we notify the client that a directories nesting level of directories for error reporting has not been set. And we propose to fix it.

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
arhiopterecs: Changed

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
arhiopterecs: Changed

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
arhiopterecs: Changed

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
arhiopterecs: Added

src/Process/Deploy/CheckEnvVarMageErrorReportDirNestingLevel.php
arhiopterecs: This description is complete.

scenario/deploy.xml
arhiopterecs: Done

arhiopterecs: Done

src/Process/Build/SetErrorReportDirNestingLevel.php
NadiyaS: @YPyltiai not applying configuration value from .magento.env.yaml goes against the behavior we have for other configurations. How do we plan to make it clear to customers? (let's say that nobody will memorise every exception for every configuration's behavior. Better to say that all what you set in .env.yaml will override existed configs - it is more clear and more expected for everyone).

YPyltiai: @NadiyaS case # 2 is very unlikely to happen - this logic just protects us from situations where we override something we shouldn't touch. It's very unlikely that someone will have that file, so in most cases behavior will be as described in # 1 - it will use value from env.yaml or follow default behavior otherwise. I think it's not the first variable with overrides we have - it just needs to be well documented.

dist/.magento.env.yaml
```diff
@@ -457,6 +457,18 @@
 #                default_setup:                                                                                       #
 #                  connection: default                                                                                #
 #######################################################################################################################
+# ERROR_REPORT_DIR_NESTING_LEVEL - Subdirectories nesting level, which indicates how many nested subdirectories will  #
```
NadiyaS: Add description that this variable is ignored if you have some file in git

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                    $reportConfigFile,
+                    $reportConfigDirNestingLevel
+                ));
+                return $this->resultFactory->success();
+            }
+
+            return $this->resultFactory->error(
+                'The directories nesting level for error reporting not set.',
+                'For set the directories nesting level use setting '
+                . '`config.report.dir_nesting_level` in the file ' . $reportConfigFile
```
NadiyaS: Please check this text with @meker12

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Validates a value of the directories nesting level
+     *
+     * @param string|int $value
+     * @return bool
+     */
+    private function validateReportDirNestingLevel($value)
+    {
+        $valueInt = (int)$value;
+        return (bool)preg_match('/^([0-9]{1,2})$/', (string)$value) && (0 <= $valueInt) && (32 >= $valueInt);
```
NadiyaS: why do you need this preg_match?

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Returns a value of the property `config.report.dir_nesting_level`
+     * from the file  <magento_root>/pub/errors/local.xml
+     *
+     * @param string $reportConfigFile
+     * @return string|null
+     * @throws FileSystemException | NotEncodableValueException
+     */
+    private function getReportConfigDirNestingLevel($reportConfigFile)
```
NadiyaS: function names have 2 type of names `ReportDirNestingLevel` and `ReportConfigDirNestingLevel`. What is the difference?

src/Config/Validator/Deploy/ReportDirNestingLevel.php
arhiopterecs: @NadiyaS  because in another case, values ​​like "some text", "2d", "d2" will be true.
See the test https://github.com/magento/ece-tools/pull/601/files#diff-f35961c9df30211dd545fc6a03d1b563R138

src/Config/Validator/Deploy/ReportDirNestingLevel.php
arhiopterecs: @NadiyaS 
`getReportConfigDirNestingLevel` It means the value obtained from the configuration file 'local.xml'

dist/.magento.env.yaml
meker12:
```suggestion
# ERROR_REPORT_DIR_NESTING_LEVEL - Subdirectories nesting level, which specifies the number of nested subdirectories to  #
```

dist/.magento.env.yaml
```diff
@@ -457,6 +457,18 @@
...
+#                                  be created in the directory `<magento_root>/var/report/` before                    #
```
meker12:
```suggestion
#                                  create in the `<magento_root>/var/report/` directory before                    #
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+     * @return void
+     */
+    private function validateEnvVarReportDirNestingLevel($envVarReportDirNestingLevel)
+    {
+        if ($this->validateReportDirNestingLevel($envVarReportDirNestingLevel)) {
+            return;
+        }
+        throw new ValidatorException(
+            sprintf(
+                'The value `%s` of the environment variable `%s` is invalid.',
```
meker12:
```suggestion
                'The value `%s` specified for the `%s` environment variable is invalid.',
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                $envVarReportDirNestingLevel,
+                Environment::ENV_MAGE_ERROR_REPORT_DIR_NESTING_LEVEL
+            ),
+            sprintf(
+                'The value of the environment variable `%s` must be an integer between 0 and 32',
```
meker12:
```suggestion
                'You must specify an integer value between 0 and 32.',
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+        if ($this->validateReportDirNestingLevel($reportConfigDirNestingLevel)) {
+            return;
+        }
+        throw new ValidatorException(
+            sprintf(
+                'The value `%s` of the property `config.report.dir_nesting_level` '
+                . 'defined in <magento_root>/pub/errors/local.xml is invalid',
+                $reportConfigDirNestingLevel
+            ),
+            'The value of the property `config.report.dir_nesting_level` must be an integer between 0 and 32'
```
meker12:
```suggestion
            'The  value for the `config.report.dir_nesting_level` property must be an integer between 0 and 32'
```

src/Step/Build/SetReportDirNestingLevel.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->stageConfig = $stageConfig;
+        $this->file = $file;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function execute()
+    {
+        $this->logger->info('Configuring directories nesting level for saving error reports');
```
meker12:
```suggestion
        $this->logger->info('Configuring directory nesting level for saving error reports');
```

dist/.magento.env.yaml
meker12:
```suggestion
# ERROR_REPORT_DIR_NESTING_LEVEL - Subdirectory nesting level, which determines the number of nested subdirectories to  #
```

src/Config/Environment.php
```diff
@@ -36,6 +36,11 @@ class Environment
     const DEFAULT_ADMIN_FIRSTNAME = 'Admin';
     const DEFAULT_ADMIN_LASTNAME = 'Username';
 
+    /**
+     * The environment variable for controlling the directories nesting level for error reporting
```
meker12:
```suggestion
     * The environment variable for controlling the directory nesting level for error reporting
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+    public function validate(): Validator\ResultInterface
+    {
+        $envVarReportDirNestingLevel = $this->environment->getEnvVarMageErrorReportDirNestingLevel();
+        $reportConfigFile = $this->configFileList->getErrorReportConfig();
+
+        try {
+            if (false !== $envVarReportDirNestingLevel) {
+                $this->validateEnvVarReportDirNestingLevel($envVarReportDirNestingLevel);
+                $this->logger->notice(sprintf(
+                    'The environment variable `%s` with the value `%s` is used to configure the directories'
```
meker12:
```suggestion
                    'The `%s` environment variable  with the value `%s` specifies a custom value for the  directory'
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                    . ' nesting level for error reporting. The environment variable has a higher priority.  Value of'
```
meker12:
```suggestion
                    . ' nesting level configured for error reporting. This value overrides the value specified in'
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                    . ' the property `config.report.dir_nesting_level` in the file %s will be ignored.',
```
meker12:
```suggestion
                    . ' the `config.report.dir_nesting_level`property in file %s , which will be ignored.',
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                    $reportConfigFile
+                ));
+                return $this->resultFactory->success();
+            }
+
+            $reportConfigDirNestingLevel = $this->getReportConfigDirNestingLevel($reportConfigFile);
+            if (null !== $reportConfigDirNestingLevel) {
+                $this->validateReportConfigDirNestingLevel($reportConfigDirNestingLevel);
+                $this->logger->notice(sprintf(
+                    'The property `config.report.dir_nesting_level` defined in the file %s with value '
```
meker12:
```suggestion
                    'The `config.report.dir_nesting_level` property defined in the file %s with value '
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+use Magento\MagentoCloud\Filesystem\Driver\File;
+use Symfony\Component\Serializer\Encoder\XmlEncoder;
+use Magento\MagentoCloud\Config\Validator;
+use Magento\MagentoCloud\Config\Validator\ValidatorException;
+use Magento\MagentoCloud\Filesystem\FileSystemException;
+use Symfony\Component\Serializer\Exception\NotEncodableValueException;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Validates a value of the directories nesting level for error reporting
```
meker12:
```suggestion
 * Validates the value specified for the directory nesting level conifgured for error reporting
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                    'The property `config.report.dir_nesting_level` defined in the file %s with value '
+                    . '`%s` and  used to configure the directories nesting level for error reporting.',
+                    $reportConfigFile,
+                    $reportConfigDirNestingLevel
+                ));
+                return $this->resultFactory->success();
+            }
+
+            return $this->resultFactory->error(
+                'The directories nesting level for error reporting not set.',
```
meker12:
```suggestion
                'The directory nesting level value for error reporting has not been configured.',
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                'For set the directories nesting level use setting '
```
meker12:
```suggestion
                'You can configure the setting using the'
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,211 @@
...
+                . '`config.report.dir_nesting_level` in the file ' . $reportConfigFile
+            );
+        } catch (FileSystemException $exception) {
+            return $this->resultFactory->error($exception->getMessage());
+        } catch (NotEncodableValueException $exception) {
+            $message = sprintf('Config of the file %s is invalid. ', $reportConfigFile);
+            $message .= $exception->getMessage();
+            return $this->resultFactory->error(
+                $message,
+                'Fix the configuration of the directories nesting level for error reporting '
```
meker12:
```suggestion
                'Fix the directory nesting level configuration for error reporting '
```

src/Step/Build/SetReportDirNestingLevel.php
```diff
@@ -0,0 +1,103 @@
...
+use Magento\MagentoCloud\Config\Stage\BuildInterface;
+use Magento\MagentoCloud\Filesystem\ConfigFileList;
+use Magento\MagentoCloud\Filesystem\Driver\File;
+use Magento\MagentoCloud\Step\StepException;
+use Magento\MagentoCloud\Step\StepInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Creates the error handling configuration file `<magento_root>/pub/errors/local.xml`
+ * with the configuration of the nesting level of directories for error reporting
```
meker12:
```suggestion
 *that specifies the directory nesting level configuration for error reporting
```

src/Step/Build/SetReportDirNestingLevel.php
```diff
@@ -0,0 +1,103 @@
...
+<config>
+    <report>
+        <dir_nesting_level>{$errorReportDirNestingLevel}</dir_nesting_level>
+    </report>
+</config> 
+XML
+            );
+            $this->logger->notice(
+                sprintf(
+                    'The file %s with property `config.report.dir_nesting_level`: `%s` was created.',
```
meker12:
```suggestion
                    'The file %s with the `config.report.dir_nesting_level` property: `%s` was created.',
```

src/Test/Unit/Config/Validator/Deploy/ReportDirNestingLevelTest.php
```diff
@@ -0,0 +1,367 @@
...
+            ->method('getEnvVarMageErrorReportDirNestingLevel')
+            ->willReturn($envVarValue);
+        $this->configFileListMock->expects($this->once())
+            ->method('getErrorReportConfig')
+            ->willReturn('<magento_root>/errors/local.xml');
+        $this->loggerMock->expects($this->once())
+            ->method('notice')
+            ->with(
+                sprintf(
+                    'The environment variable `MAGE_ERROR_REPORT_DIR_NESTING_LEVEL` with the value `%s`'
```
meker12:
```suggestion
                    'The `MAGE_ERROR_REPORT_DIR_NESTING_LEVEL` environment variable with the value `%s`'
```

src/Config/Validator/Deploy/ReportDirNestingLevel.php
NadiyaS: doesn't `is_numeric()` work for it?

dist/.magento.env.yaml
arhiopterecs: changes were applied

src/Config/Validator/Deploy/ReportDirNestingLevel.php
arhiopterecs: Done

dist/.magento.env.yaml
arhiopterecs: Done

src/Config/Environment.php
```diff
@@ -113,6 +118,16 @@ public function getEnvVarName(string $name): string
         return $this->systemConfig->get($name);
     }
 
+    /**
+     * Get a value of the environment variable MAGE_ERROR_REPORT_DIR_NESTING_LEVEL.
+     *
+     * @return int|null|string
```
oshmyheliuk: Method getEnv has more return types. `@return array|string|int|null|bool`

Maybe is better to convert the result of this function to integer before return?

arhiopterecs: @oshmyheliuk 
No need. We present only information about the value of this environment variable. And we show this value as it is.

oshmyheliuk: Ok, but in this case, return types should be the same as in the `getEnv` method.

patches.json
```diff
@@ -180,6 +180,18 @@
...
       "2.3.2 - 2.3.3": "MAGECLOUD-4071__terminate_consumers_if_the_queue_is_empty__2.3.2.patch"
+    },
+    "Reduce q-ty of error report files": {
+      "2.1.4 - 2.1.7": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.4.patch",
+      "2.1.8 - 2.1.16": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.8.patch",
+      "~2.1.17": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.1.17.patch",
+      "2.2.0 - 2.2.4": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.2.0.patch",
+      "2.2.5": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.2.5.patch",
+      "2.2.6 - 2.2.7": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.2.6.patch",
+      "~2.2.8": "MAGECLOUD-3392__reduce_q-ty_of_error_report_files__2.2.8.patch",
```
shiftedreality, line 191: Don't you plan to deliver changes to core 2.1.18 and 2.2.9?

arhiopterecs: @shiftedreality No, because Magento 2 1.x is no longer supported
For 2.2.x only security PRs are delivered

src/Config/Environment.php
arhiopterecs: Done

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function validate(): Validator\ResultInterface
+    {
+        $envVarValue = $this->environment->getEnvVarMageErrorReportDirNestingLevel();
+        $reportConfigFile = $this->configFileList->getErrorReportConfig();
+
+        try {
+            if (false !== $envVarValue) {
```
NadiyaS: so, you decided to leave validation on the deploy phase which is not related to cloud?

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    /**
+     * Returns a value of the property `config.report.dir_nesting_level`
+     * from the file  <magento_root>/pub/errors/local.xml
+     *
+     * @param string $file
+     * @return int|string|null
+     * @throws FileSystemException | NotEncodableValueException
```
oshmyheliuk: It looks like here is wrong @throws format.
Should be:
```
@throws FileSystemException
@throws NotEncodableValueException
```
https://docs.phpdoc.org/references/phpdoc/tags/throws.html
`It is also RECOMMENDED that this tag occurs for every occurrence of an exception`

oshmyheliuk (dismissed):

src/Config/Validator/Deploy/ReportDirNestingLevel.php
arhiopterecs: Done

arhiopterecs: Done

src/Config/Validator/Deploy/ReportDirNestingLevel.php
arhiopterecs: @NadiyaS I removed the code that was responsible for validation, leaving only a check for the existence of the environment variable `MAGE_ERROR_REPORT_DIR_NESTING_LEVEL`. This is necessary to evaluate the case when the report directory nesting level not setting is not used.

src/Config/Validator/Deploy/ReportDirNestingLevel.php
```diff
@@ -0,0 +1,120 @@
...
+                . ' in the file ' . $reportConfigFile
+            );
+        } catch (FileSystemException $exception) {
+            return $this->resultFactory->error($exception->getMessage());
+        } catch (NotEncodableValueException $exception) {
+            $message = sprintf('Config of the file %s is invalid. ', $reportConfigFile);
+            $message .= $exception->getMessage();
+            return $this->resultFactory->error(
+                $message,
+                'Fix the directory nesting level configuration for error reporting in the file '
```
NadiyaS, line 97: As I see this error can be caught if there is some problem with decoding xml file. And the report dir level can be correct, so this suggestion is not always true.

NadiyaS (approved):

oshmyheliuk (approved):

andriyShevtsov:
QA approved

PR #620, opened 2019-10-08
MAGECLOUD-4109:  Introduce extensibility mechanism based on Dependency Injection

oshmyheliuk:
### Description

https://magento2.atlassian.net/browse/MAGECLOUD-4109
https://magento2.atlassian.net/browse/MAGECLOUD-4110

Example of how to extend ece-tools:
https://github.com/oshmyheliuk/ece-tools-extend

### Manual testing scenarios

https://jira.corp.magento.com/browse/MAGECLOUD-140

### Contribution checklist


YPyltiai (dismissed):
Priorities definition flow & increment (100) approved.
Please proceed on code review.

config/services.xml
```diff
@@ -32,6 +32,7 @@
         <service id="Magento\MagentoCloud\Step\Deploy\InstallUpdate\Install" autowire="false" />
         <service id="Magento\MagentoCloud\Step\Deploy\InstallUpdate\Update" autowire="false" />
         <service id="Magento\MagentoCloud\Step\Deploy\PreDeploy" autowire="false" />
+        <service id="Magento\MagentoCloud\Step\DummyStep" autowire="false" />
```
shiftedreality: This should be removed

src/ExtensionRegistrar.php
```diff
@@ -0,0 +1,57 @@
...
+    public static function register($componentName, $path)
+    {
+        if (isset(self::$paths[$componentName])) {
+            throw new \LogicException(sprintf(
+                "%s from %s has been already defined in %s",
+                $componentName,
+                $path,
+                self::$paths[$componentName]
+            ));
+        } else {
```
shiftedreality, line 37: `else` statement can be avoided as `if` throws exception

src/ExtensionRegistrar.php
```diff
@@ -0,0 +1,57 @@
...
+            self::$paths[$componentName] = str_replace('\\', '/', $path);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
```
shiftedreality: Remove curly brackets

src/ExtensionRegistrar.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static function getPaths()
+    {
+        return self::$paths;
+    }
+
+    /**
+     * {@inheritdoc}
```
shiftedreality: Remove curly brackets

src/ExtensionRegistrar.php
```diff
@@ -0,0 +1,57 @@
...
+            ));
+        } else {
+            self::$paths[$componentName] = str_replace('\\', '/', $path);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getPaths()
```
shiftedreality: Add return type

src/ExtensionRegistrar.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Sets the location of an extension.
+     *
+     * @param string $componentName Fully-qualified component name
+     * @param string $path Absolute file path to the component
+     * @throws \LogicException
+     * @return void
+     */
+    public static function register($componentName, $path)
```
shiftedreality: Add return types including `void` everywehere

src/Scenario/Collector/Step.php
```diff
@@ -0,0 +1,143 @@
...
+                    'Argument(s) "%s" are missed from item "%s"',
+                    implode(', ', $missedArgs),
+                    $itemName
+                ));
+            }
+
+            $newItem = [
+                'name' => $itemName,
+                'xsi:type' => $item['@xsi:type'],
+                'priority' => $item['@priority'] ?? self::DEFAULT_PRIORITY,
```
shiftedreality, line 128: Priority should be required

src/Scenario/Resolver.php
```diff
@@ -40,11 +50,20 @@ public function resolve(array $scenarios): array
...
 
+        $this->sorter->sortScenarios($scenarios);
+
         foreach ($scenarios as $step) {
-            $instance = $this->container->create(
-                $step['type'],
-                $this->resolveParams($step['arguments'] ?? [], $step['name'])
-            );
+            if (!$step['type']) {
+                $instance = $this->container->create(DummyStep::class, [
```
shiftedreality: Why do we have `DummyStep`?

shiftedreality (changes requested):

config/services.xml
oshmyheliuk: No, we need some Step which will be logging about the skipped steps.
Maybe we need to change the name of this class.

shiftedreality: Please revert step remove functionality.
It must use `remove="true"` attribute instead

src/Scenario/Resolver.php
oshmyheliuk: To add information into cloud.log about skipped steps.

src/Scenario/Collector/Step.php
oshmyheliuk: Priority is required for steps, but not required for items as not all item tags need priority,
for example:
```
<item name="500" xsi:type="array">
                    <item name="database-configuration" xsi:type="object">Magento\MagentoCloud\Config\Validator\Deploy\DatabaseConfiguration</item>
                    <item name="search-configuration" xsi:type="object">Magento\MagentoCloud\Config\Validator\Deploy\SearchConfiguration</item>
                </item>
```
in this case `<item name="500" xsi:type="array">` doesn't need a priority

oshmyheliuk:
@YPyltiai Are you okay with such mechanism of step skipping:
https://github.com/oshmyheliuk/ece-tools-extend/blob/1455b5c864e8c5d4d07bf2a81cc055e0dd491e88/scenario/extend-deploy.xml#L14

shiftedreality:
@oshmyheliuk this is a very intuitive way that will confuse most users. Skipping arguments should mean that they're optional, not removed.

The correct way is using attribute like `remove="true"`

src/ExtensionRegistrar.php
oshmyheliuk: added

src/ExtensionRegistrar.php
oshmyheliuk: added

src/ExtensionRegistrar.php
oshmyheliuk: tag removed

config/services.xml
oshmyheliuk: Functionality doesn't need to be reverted it will be updated If we will decide to use attribute instead.

src/Scenario/Resolver.php
oshmyheliuk: We can remove this Class but in this case, cloud.log won't contain information about skipped steps. I think this information is important for us in case of escalation.

oshmyheliuk:
@shiftedreality I don't think that there are correct/not-correct ways it's just different implementation.

YPyltiai:
@shiftedreality , @oshmyheliuk , I think attributes may be more clear long-term, since you explicitly define that it should be skipped/removed. Also I would not remove path to code itself just to make sure users always have reference/path to what exactly they disabled.

oshmyheliuk:
Step skipping was changed from empty tag value to attribute `skip`.
@shiftedreality Please review again

config/services.xml
shiftedreality: Let's call this class, `SkipStep`, it's not `dummy`

shiftedreality (changes requested):

config/services.xml
oshmyheliuk: Ok, the class was renamed.

shiftedreality (approved):

NadiyaS (approved):

mveeramneni:
QA Approved

PR #658, opened 2019-12-19
MAGECLOUD-4929: Programmatically generate .magento.env.yaml.dist - Schema declaration and generation

shiftedreality:

1. https://magento2.atlassian.net/browse/MAGECLOUD-4025
2. https://magento2.atlassian.net/browse/MAGECLOUD-4929

### Manual testing scenarios

1. Run `./vendor/bin/ece-tools schema:generate`
2. Validate output in `.magento.env.md` file

### Contribution checklist


config/schema.yaml
```diff
@@ -0,0 +1,128 @@
...
+  default:
+    build: ''
+    deploy: ''
+  examples:
+    - stage:
+        build:
+          VERBOSE_COMMANDS: '-vv'
+    - stage:
+        deploy:
+          VERBOSE_COMMANDS: '-vv-'
```
oshmyheliuk: redundant dash at the end

src/Config/Schema/Validator/CompressionLevel.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Magento\MagentoCloud\Config\Schema\Validator;
+
+use Magento\MagentoCloud\Config\Validator\ResultFactory;
+use Magento\MagentoCloud\Config\Validator\ResultInterface;
+
+/**
+ * Validates the compression level value
+ */
+class CompressionLevel implements ValidatorInterface
```
oshmyheliuk, line 16: Maybe better to create one `RangeValidator` for all validators which need to check if the value in the given range and configure range in .yaml file
Now we have at least two variables with such validations.

config/schema.yaml
shiftedreality: Fixed

src/Config/Schema/Validator/CompressionLevel.php
shiftedreality: It's kindy tricky to pass the range to such class, we can have this as possible improvement

michaelcasey316 (dismissed):

michaelcasey316 (dismissed):

src/App/Container.php
```diff
@@ -124,4 +130,19 @@ public function create(string $abstract, array $params = [])
...
         return new $abstract(...array_values($params));
     }
+
+    /**
+     * @param string $toolsBasePath
+     * @param string $magentoBasePath
+     * @return Container
+     * @throws ContainerException
+     */
+    public static function getInstance(string $toolsBasePath, string $magentoBasePath): Container
```
oshmyheliuk: Why do you need this method? I see only one usage in tests, there can be used `new Container()` as well.

config/schema.yaml
```diff
@@ -0,0 +1,340 @@
...
+      global:
+        SCD_ON_DEMAND: true
+MIN_LOGGING_LEVEL:
+  description: Use to override the minimum logging level for all output streams without making changes to the code.
+    This helps to improve troubleshooting problems with deployment. For example, if your deployment fails,
+    you can use this variable to increase the logging granularity globally. Doesn't affect log level for the file output.
+    See Set up notifications—Log levels.
+  type: string
+  allowed:
+    - ''
```
oshmyheliuk, line 181: Do we need to allow an empty value? If an option is added to .magento.env.yaml it shouldn't be empty.

shiftedreality: Makes sense, fixed

src/App/Container.php
shiftedreality: It dramatically improves the integration test performce

config/schema.yaml
shiftedreality: But actually you will lose possibility to overwrite it per stage

config/schema.yaml
```diff
@@ -0,0 +1,340 @@
+# Stage variables
```
oshmyheliuk: option ERROR_REPORT_DIR_NESTING_LEVEL is missed in this file

oshmyheliuk, line 1: option ERROR_REPORT_DIR_NESTING_LEVEL is missed in this file
```BuildInterface::VAR_ERROR_REPORT_DIR_NESTING_LEVEL```

Comment (shiftedreality) [reply config/schema.yaml]: There are bunch of them missed, will be added in next task

Comment (oshmyheliuk) [reply config/schema.yaml:1]: Plus some other options are missed:
ELASTICSUITE_CONFIGURATION
CACHE_CONFIGURATION
DATABASE_CONFIGURATION
...

File: src/Config/Schema.php [hunk 6]
```diff
@@ -61,291 +89,44 @@ public function getDefaults(string $stage): array
...
-                self::SCHEMA_TYPE => ['boolean'],
-                self::SCHEMA_STAGE => [
-                    StageConfigInterface::STAGE_GLOBAL,
-                    StageConfigInterface::STAGE_DEPLOY
-                ],
-                self::SCHEMA_DEFAULT_VALUE => [
-                    StageConfigInterface::STAGE_DEPLOY => true,
-                ],
-            ],
             DeployInterface::VAR_SEARCH_CONFIGURATION => [
```
oshmyheliuk, line 127: Why do we keep some option configuration here? It should be moved to `.yaml` file

src/App/Container.php
oshmyheliuk: So, it will make sense to make `__construct` private and change where we use `new Container()` to `Container::getInstance`

shiftedreality:
All missing variables will be moved in the scope of https://magento2.atlassian.net/browse/MAGECLOUD-4970

src/Config/Schema.php
shiftedreality: Will be moved in next task

config/schema.yaml
shiftedreality: Will be moved in next task

src/App/Container.php
shiftedreality: Making constructor to be private does not give and advantage for generic scenario

shiftedreality:
@YPyltiai @hguthrie I've updated generated `.MD` file to be like this:

# VERBOSE_COMMANDS

Adjust the output verbosity level of Magento commands.

|Attribute|Values|
|---|---|
|Type|string|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|
|Allowed values|-v, -vv, -vvv|

## Examples

```yaml
  stage:
      build:
          VERBOSE_COMMANDS: '-vv'

```

```yaml
  stage:
      deploy:
          VERBOSE_COMMANDS: '-vvv'

```

# SCD_COMPRESSION_LEVEL

Gzip compression level (0 to 9) to use when compressing static content; 0 disables compression.

|Attribute|Values|
|---|---|
|Type|integer|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|

## Examples

```yaml
  stage:
      build:
          SCD_COMPRESSION_LEVEL: 8

```

# SCD_COMPRESSION_TIMEOUT

The maximum time (in seconds) allowed for compressing static content.

|Attribute|Values|
|---|---|
|Type|integer|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|

## Examples

```yaml
  stage:
      build:
          SCD_COMPRESSION_TIMEOUT: 900

```

# SCD_STRATEGY

Customize the deployment strategy for static content. The value "standard" will deploy all static
view files for all packages, "quick" will minimize deployment time, and "compact" will conserve
space on the server. Compact is not compatible with multi-threaded deployment and will set the value
of `SCD_THREADS` to 1. The default value is "quick".

|Attribute|Values|
|---|---|
|Type|string|
|Magento version|\>=2.2.0|
|Stages|global, build, deploy|
|Allowed values|compact, quick, standard|

## Examples

```yaml
  stage:
      build:
          SCD_STRATEGY: compact

```

# SCD_THREADS

The number of threads for static content deployment

|Attribute|Values|
|---|---|
|Type|integer|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|

## Examples

```yaml
  stage:
      global:
          SCD_THREADS: 2

```

# SCD_MAX_EXECUTION_TIME

Allows you to increase the maximum expected execution time for the static content deploy command. By
default, Magento Commerce sets the maximum expected execution to 400 seconds, but in some scenarios
you might need more time to complete the static content deployment for a Cloud project.

|Attribute|Values|
|---|---|
|Type|integer|
|Magento version|\>=2.2.0|
|Stages|global, build, deploy|

## Examples

```yaml
  stage:
      global:
          SCD_MAX_EXECUTION_TIME: 3600

```

# SCD_MATRIX

Allows you to configure multiple locales per theme as long as the theme is not excluded using the
SCD_EXCLUDE_THEMES variable during deployment. This is ideal if you want to speed up thedeployment
process by reducing the amount of unnecessary theme files. For example, you can deploy the
magento/backend theme in English and a custom theme in other languages.

|Attribute|Values|
|---|---|
|Type|array|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|

## Examples

The following example deploys the magento/backend theme with three locales

```yaml
  stage:
      deploy:
          SCD_MATRIX:
              magento/backend:
                  language:
                      - en_US
                      - fr_FR
                      - af_ZA

```

Also, you can choose to not deploy a theme

```yaml
  stage:
      deploy:
          SCD_MATRIX:
              magento/backend: {  }

```

# SKIP_SCD

Skips static content deployment during the build/deploy phase.

|Attribute|Values|
|---|---|
|Type|boolean|
|Magento version|\>=2.1.14|
|Stages|global, build, deploy|

## Examples

```yaml
  stage:
      build:
          SKIP_SCD: true

```

# SKIP_HTML_MINIFICATION

Skip copying the static view files in the var/view_preprocessed directory to reduce downtime when
deploying to the Staging and Production environments and generates minified HTML when requested.

|Attribute|Values|
|---|---|
|Type|boolean|
|Magento version|\>=2.1.14|
|Stages|global|

## Examples

```yaml
  stage:
      global:
          SKIP_HTML_MINIFICATION: true

```

Please let me know if this looks good for you

src/App/Container.php
oshmyheliuk: It's not right when the class has static `getInstance` method and can be created via `new` and as a result, it will be two different instances of the same class.

oshmyheliuk: If you need this only for tests better to move `getInstance` method to some factory in the "Test" namespace

src/Test/Unit/Config/Schema/_files/.magento.env.md
```diff
@@ -0,0 +1,20 @@
+# ENV_VARIABLE
```
hguthrie: In DevDocs, the heading 1 (single hash) is reserved for the page title. Not sure if that rule should apply here, but if there is a page title, you may want to consider making this a heading 2.

src/Test/Unit/Config/Schema/_files/.magento.env.md
```diff
@@ -0,0 +1,20 @@
...
+
+Some description
+
+|Attribute|Values|
+|---|---|
+|Type|string|
+|Magento version|\>=2.1.24|
+|Stages|global|
+
+## Examples
```
hguthrie: Subsequently, this would be a heading 3.

hguthrie (commented):
The markdown looks nice @shiftedreality . Other than the heading levels, the only other thing I could comment about would be the order in which these variables display. Is there an order? Would you recommend it be based on common use or alphabetizing...or will it matter?

src/App/Container.php
shiftedreality: Added `Test/Integration/Container` with this method

oshmyheliuk (dismissed):

src/Test/Unit/Config/Schema/_files/.magento.env.md
shiftedreality: Fixed

src/Test/Unit/Config/Schema/_files/.magento.env.md
shiftedreality: Fixed

YPyltiai:
@shiftedreality markdown looks good. The only thing - why `SCD_COMPRESSION_LEVEL` has no allowed values? You can put 0-9 there, where 0 disables it. So this should be there I assume

shiftedreality:
@YPyltiai I will add this to the scope of next task to finalize moving of env variables to schema

oshmyheliuk (approved):

shiftedreality:
Tasks to continue work:
- https://magento2.atlassian.net/browse/MAGECLOUD-4970
- https://magento2.atlassian.net/browse/MAGECLOUD-4988

cc: @andriyShevtsov

andriyShevtsov:
QA Approved. But this task isn't fully ready yet because we need to add description for all variables and fix templates

michaelcasey316 (approved):

PR #660, opened 2020-01-08
MAGECLOUD-3011: Add Baler Support

bbatsche:
### Description

Adds a new build configuration variable (`SCD_USE_BALER`) that will instruct ECE Tools to run `baler` as part of the SCD on build process.

1. [MAGECLOUD-4592](https://magento2.atlassian.net/browse/MAGECLOUD-4592)
1. [MAGECLOUD-4593](https://magento2.atlassian.net/browse/MAGECLOUD-4593) 

### Manual testing scenarios

[MAGECLOUD-147](https://jira.corp.magento.com/browse/MAGECLOUD-147)

### Contribution checklist

releasenotes
Added the [SCD_RUN_BALER](https://devdocs.magento.com/cloud/env/variables-build.html#scd_use_baler) environment variable to enable javascript bundling during the build process.


YPyltiai:
@bbatsche , if the dependency is not installed - what would happen then? Should this be covered by validation?

EDIT: looks like I missed it in code and you already have it.

scenario/build/generate.xml
```diff
@@ -42,4 +42,13 @@
             </argument>
         </arguments>
     </step>
+    <step name="run-baler" type="Magento\MagentoCloud\Step\Build\RunBaler" priority="1300">
+        <arguments>
+            <argument name="logger" xsi:type="object">Psr\Log\LoggerInterface</argument>
```
oshmyheliuk, line 47: Dependencies are autoloading automatically. You don't need to describe arguments, except you want to use a custom implementation of some interface or configure some static-type argument.

oshmyheliuk: Please ignore my comment above, I see that arguments are required in your case.

src/Config/Validator/Build/BalerSupport.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $magentoConfig = $this->configReader->read();
+
+        if (!isset($magentoConfig['modules']['Magento_Baler']) || !$magentoConfig['modules']['Magento_Baler']) {
+            $errors[] = 'The Magento_Baler module is not installed or is disabled.';
+        }
+
+        $jsConfig = $magentoConfig['system']['default']['dev']['js'] ?? [];
+
+        if (!isset($jsConfig['enable_baler_js_bundling']) || !$jsConfig['enable_baler_js_bundling']) {
```
shiftedreality, line 68: Can't it be enabled via DB?

src/Config/Validator/Build/BalerSupport.php
```diff
@@ -0,0 +1,85 @@
...
+            $errors[] = 'The Magento config dev/js/enable_baler_js_bundling must be enabled in app/etc/config.php.';
+        }
+
+        foreach (['minify_files', 'enable_js_bundling', 'merge_files'] as $configKey) {
+            if (!isset($jsConfig[$configKey]) || $jsConfig[$configKey]) {
+                $errors[] = sprintf('The Magento config dev/js/%s must be disabled in app/etc/config.php.', $configKey);
+            }
+        }
+
+        return empty($errors)
```
shiftedreality: `empty` call should be avoided

src/Step/Build/RunBaler.php
```diff
@@ -0,0 +1,97 @@
...
+     * @var ValidatorInterface
+     */
+    private $validator;
+
+    /**
+     * @var ShellInterface
+     */
+    private $shell;
+
+    public function __construct(
```
shiftedreality, line 54: PHPdoc is missing

src/Step/Build/RunBaler.php
```diff
@@ -0,0 +1,97 @@
...
+        ShellInterface $shell
+    ) {
+        $this->logger = $logger;
+        $this->buildConfig = $buildConfig;
+        $this->flagManager = $flagManager;
+        $this->validator = $validator;
+        $this->shell = $shell;
+    }
+
+    public function execute()
```
shiftedreality, line 71: Add inheritdoc

src/Step/Build/RunBaler.php
```diff
@@ -0,0 +1,97 @@
...
+        LoggerInterface $logger,
+        BuildInterface $buildConfig,
+        Flag\Manager $flagManager,
+        ValidatorInterface $validator,
+        ShellInterface $shell
+    ) {
+        $this->logger = $logger;
+        $this->buildConfig = $buildConfig;
+        $this->flagManager = $flagManager;
+        $this->validator = $validator;
```
shiftedreality, line 64: You must use factory instead of passing implementation via DI

shiftedreality (changes requested):

src/Config/Validator/Build/BalerSupport.php
```diff
@@ -0,0 +1,85 @@
...
+            $errors[] = 'The Magento_Baler module is not installed or is disabled.';
+        }
+
+        $jsConfig = $magentoConfig['system']['default']['dev']['js'] ?? [];
+
+        if (!isset($jsConfig['enable_baler_js_bundling']) || !$jsConfig['enable_baler_js_bundling']) {
+            $errors[] = 'The Magento config dev/js/enable_baler_js_bundling must be enabled in app/etc/config.php.';
+        }
+
+        foreach (['minify_files', 'enable_js_bundling', 'merge_files'] as $configKey) {
```
oshmyheliuk, line 72: Are these options 'minify_files', 'enable_js_bundling', 'merge_files' are enabled by default in Magento?
If they aren't enabled by default, next check `!isset($jsConfig[$configKey])` will mean that config is disabled and it won't be an error.

src/Config/Validator/Build/BalerSupport.php
```diff
@@ -0,0 +1,85 @@
...
+    public function validate(): Validator\ResultInterface
+    {
+        $errors = [];
+
+        if ($this->finder->find('baler') === null) {
+            $errors[] = 'Path to baler executable could not be found.'
+                . ' The Node package may not be installed or may not be linked.';
+        }
+
+        $magentoConfig = $this->configReader->read();
```
oshmyheliuk, line 60: You are reading configuration only from `config.php` file but it can be stored in the database or env variables as well. For modules it's ok, but for system config, you should use `bin/magento config:show` command.

bbatsche: Will `config:show` work without the DB though? One of the agreements Jeff and I came to was that we would only support running Baler during build, not the deployment phase, as an added incentive for users to do SCD on build.

src/Config/Validator/Build/BalerSupport.php
bbatsche: I believe these values are `'0'` when the config file is dumped by default, however part of the intent is that the user should _know_ these options are being disabled, not just rely on the defaults from Magento, so that we ensure future compatibility with Baler

src/Step/Build/RunBaler.php
bbatsche: I'm not entirely sure how you intend for me to do that? I followed the same paradigm that `Step\Build\DeployStaticContent` did in `generate.xml`.

src/Config/Validator/Build/BalerSupport.php
oshmyheliuk: I think no, my bad I forgot that we don't have access to the database on the Build phase.
So this functionality requires to have dumped configuration into config.php.

oshmyheliuk (dismissed):

shiftedreality (dismissed):

shiftedreality (approved):

oshmyheliuk (approved):

mveeramneni:
QA Approved

PR #665, opened 2020-01-14
MAGECLOUD-4970: Move all environment variables to schema config

oshmyheliuk:
### Description

https://magento2.atlassian.net/browse/MAGECLOUD-4970

### Manual testing scenarios

1. Run `php bin/ece-tools schema:generate` comnmand
2. Check that all configuration options exist in the dumped file `.magento.env.md`

### Contribution checklist


config/schema.yaml
```diff
@@ -1,6 +1,6 @@
 # Stage variables
 VERBOSE_COMMANDS:
-  description: Adjust the output verbosity level of Magento commands.
+  description: "Adjust the output verbosity level of Magento commands."
```
shiftedreality: There is no need in quotes - https://yaml.org/spec/1.2/spec.html#id2760844

shiftedreality: http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html

oshmyheliuk: Some description messages contain colon symbol `:` and without quotes, `:<space>` won't work.
I just changed all description messages to one style.

shiftedreality: This will be lost on the dump in any case, please remove colons where they’re not needed

oshmyheliuk: Colon won't be lost on the dump. We have a lot of variables with such construction `_merge: true` in the description. Also, there is one message which requires newlines, without quotes it won't work too.

shiftedreality: Quotes will be lost, not colons. Yaml Dump class by default follows yaml spec and won’t dump quotes

oshmyheliuk: Yes, I understand that quotes will be lost. But as I mentioned some descriptions won't work without quotes, and I just changed all descriptions to one format.

config/schema.yaml
```diff
@@ -281,6 +283,299 @@ CLEAN_STATIC_FILES:
...
+    - deploy
+  default:
+    deploy: []
+  examples:
+    - comment: Default value depends on template configuration of ElasticSearch (i.e. default number of replicas and shards)
+        and ElasticSearch host.
+      stage:
+        deploy:
+          ELASTICSUITE_CONFIGURATION:
+            _merge: false if true configuration merging with default"
```
billygilbert: If this is an example, it should just be true/false.  Also looks like an extra "

config/schema.yaml
```diff
@@ -281,6 +283,299 @@ CLEAN_STATIC_FILES:
...
+        deploy:
+          CONSUMERS_WAIT_FOR_MAX_MESSAGES: true
+ENABLE_GOOGLE_ANALYTICS:
+  description: Prevents the disabling of Google Analytics.  By default, Google Analytics is enabled only
+    on Production environments. Set this value to true to enable Google Analytics on other environments as well.
+    Google Analytics is always enabled in Production.
+  type: boolean
+  stages:
+    - deploy
+    - global
```
billygilbert, line 529: Why do we note global here and not other places (example line 508)

config/schema.yaml
```diff
@@ -281,6 +283,299 @@ CLEAN_STATIC_FILES:
...
+    - stage:
+        deploy:
+          GENERATED_CODE_SYMLINK: false
+WARM_UP_PAGES:
+  description: Customize the list of pages used to preload the cache in the post_deploy stage.
+  type: array
+  stages:
+    - post-deploy
+  default:
+    post-deploy: [""]
```
billygilbert, line 554: should the "" be there?  Different than 560 which is a similar config.

config/schema.yaml
```diff
@@ -281,6 +283,299 @@ CLEAN_STATIC_FILES:
...
+  examples:
+    - stage:
+        post-deploy:
+          WARM_UP_PAGES:
+            - "index.php"
+            - "index.php/customer/account/create"
```
billygilbert, line 560: Should we have an example here showing full URL? IE https://example.com/checkout/cart

config/schema.yaml
```diff
@@ -338,3 +633,21 @@ ENV_ENVIRONMENT:
     - system:
         variables:
           ENV_ENVIRONMENT: MAGENTO_CLOUD_ENVIRONMENT
+
+# Dev variables
```
billygilbert, line 645: I believe these variables are not technically supported outside of our internal teams, I don't know if we want to document them.  @YPyltiai, thoughts?

billygilbert (changes requested):

config/schema.yaml
shiftedreality: String, where allowed value is boolean, may be confusing, possibly better to move to comment section

config/schema.yaml
```diff
@@ -281,6 +283,299 @@ CLEAN_STATIC_FILES:
...
+  type: array
+  stages:
+    - deploy
+  default:
+    deploy: []
+  examples:
+    - stage:
+        deploy:
+          CACHE_CONFIGURATION:
+            _merge: false if true configuration merging with default
```
shiftedreality: Same here, these options should be boolean

src/Command/GenerateSchema.php
```diff
@@ -82,9 +91,13 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         $data = $this->schema->getSchema();
 
+        $this->logger->info('Starting schema dist file generation');
+
         $this->file->filePutContents(
             $this->fileList->getEnvDistConfig(),
             $this->formatter->format($data)
         );
+
+        $this->logger->info(sprintf('Dist file was successfully generated: %s', $this->fileList->getEnvDistConfig()));
```
shiftedreality: Logging in command classes does not have much value, Command output should be used instead

src/Config/Schema/Formatter.php
```diff
@@ -39,14 +39,18 @@ public function format(array $data): string
         $text = '';
 
         foreach ($data as $key => $item) {
+            if (isset($item[Schema::SCHEMA_DUMP]) && !$item[Schema::SCHEMA_DUMP]) {
```
shiftedreality: `isset()` should be eliminated in favor to `!empty()` only

shiftedreality (changes requested):

config/schema.yaml
oshmyheliuk: Moved from old `Schema.php` file without changes to avoid BIC

config/schema.yaml
oshmyheliuk: Yes, this is a specific of this particular configuration. It moved from `Schema.php` too.

config/schema.yaml
oshmyheliuk: Dev variables won't dump to the documentation if they have an option: `dump: false`.
But we need them in schema.yaml for the ability to use them in the code.

src/Config/Schema/Formatter.php
oshmyheliuk: `!empty` doesn't work in this case. We need to check if option configured and it `false`

config/schema.yaml
oshmyheliuk: Example added

config/schema.yaml
oshmyheliuk: Moved `if true configuration merging with default` to the example comment section

src/Config/Schema/Formatter.php
oshmyheliuk: done

config/schema.yaml
```diff
@@ -311,10 +311,11 @@ SEARCH_CONFIGURATION:
   default:
     deploy: []
   examples:
-    - stage:
+    - comment: If `_merge` is true configuration merging with default
```
YPyltiai: `configuration merging with default` -> `configuration is being merged with the default`

config/schema.yaml
YPyltiai: If they are not going to Docs - Sasha's explanation should work.

src/Command/GenerateSchema.php
oshmyheliuk: Changed

config/schema.yaml
oshmyheliuk: Thanks, changed.

billygilbert (dismissed):

shiftedreality (dismissed):

billygilbert (approved):

shiftedreality (approved):

andriyShevtsov:
QA approved

Issue #669, opened 2020-01-23
Remove hard dependency on redis-cli


Flyingmana:
### Preconditions

1. Have an Environment / Docker container, which does not ship with the redis-cli

### Steps to reproduce

1.  Have an Environment / Docker container, which does not ship with the redis-cli
2. use ece-tools
3. failing because redis-cli is not available to execute flushdb

### Expected result

1. Instead of using the redis-cli, the same should be used, which is also used in Magento.
(Credis?) or the underlying php Redis Extension.

### Actual result

1.  failing because redis-cli is not available to execute flushdb
```
The command "redis-cli -h redis -p 6379 -n 0 flushdb" failed.
Exit Code: 127(Command not found)
```

### References
https://github.com/magento/ece-tools/issues/409
https://github.com/magento/ece-tools/pull/410

### Suggestions

Instead of just switching it out, would it be more acceptable if we make it configurable and just add it as optional alternative?

shiftedreality:
We could potentially use credis extension

YPyltiai:
Looks like @Flyingmana started working on it in #670

PR #670, opened 2020-01-23
MAGECLOUD-5147: Remove hard dependency on redis-cli

Flyingmana:
see #669 

in Draft state to see whats needed to get the Tests be green


shiftedreality:
@Flyingmana thanks for your contribution.

Please let us know once it ready for review.

src/Step/Deploy/PreDeploy/CleanRedisCache.php
```diff
@@ -74,12 +74,10 @@ public function execute()
             }
             $redisConfig = $cacheConfig['backend_options'];
             $this->logger->info("Clearing redis cache: $cacheType");
-            $cmd = 'redis-cli';
-            $cmd .= isset($redisConfig['server']) ? ' -h ' . $redisConfig['server'] : '';
-            $cmd .= isset($redisConfig['port']) ? ' -p ' . $redisConfig['port'] : '';
-            $cmd .= isset($redisConfig['database']) ? ' -n ' . $redisConfig['database'] : '';
-            $cmd .= ' flushdb';
-            $this->shell->execute($cmd);
+            $redisClient = new \Redis();
```
kandy: Maybe will be even better to use PHP implementation of Redis client like colinmollenhour/credis.

Flyingmana: If its ok to add this dependency, I would absolutely support it and change this PR accordingly.

YPyltiai: @kandy ^ please review. I assume it's okay given Magento core anyway uses it? Not sure if we even need to explicitly add the dependency here or just let core decide. I would assume we want some fallback in this case too?

kandy: Magento already have dependencies on `colinmollenhour/credis` from 1.6 - 1.10
So, I think it's ok to add it to ece-tools as well as  `"colinmollenhour/credis": "1.*"`

shiftedreality:
Internal ticket: https://magento2.atlassian.net/browse/MAGECLOUD-3252

Flyingmana:
will finish this next week

Flyingmana:
I did already sign the CLA, currently checking what did go wrong there.

Regarding the changes, Iam not sure how to align the CleanRedisCacheTest ( https://github.com/magento/ece-tools/blob/develop/src/Test/Unit/Step/Deploy/PreDeploy/CleanRedisCacheTest.php )
Iam already surprised its not causing the build to fail. 
As the Test would just cover the $cacheConfig['backend_options'] are correctly aligned to the Credis constructor arguments, Iam not sure if the test is even needed anymore.
Any recommendations, wishes how this should be solved?

composer.json
```diff
@@ -15,6 +15,7 @@
     "ext-PDO": "*",
     "ext-json": "*",
     "ext-sockets": "*",
+    "colinmollenhour/credis": "^1.6.0",
```
YPyltiai, line 18: Not sure on version. I would make it more loose and let Core decide which one to pick. As @kandy suggested, `^1.0.0`  or `~1.0` may work.

src/Step/Deploy/PreDeploy/CleanRedisCache.php
```diff
@@ -74,12 +75,15 @@ public function execute()
...
             $redisConfig = $cacheConfig['backend_options'];
             $this->logger->info("Clearing redis cache: $cacheType");
-            $cmd = 'redis-cli';
-            $cmd .= isset($redisConfig['server']) ? ' -h ' . $redisConfig['server'] : '';
-            $cmd .= isset($redisConfig['port']) ? ' -p ' . $redisConfig['port'] : '';
-            $cmd .= isset($redisConfig['database']) ? ' -n ' . $redisConfig['database'] : '';
-            $cmd .= ' flushdb';
-            $this->shell->execute($cmd);
+            $redisClient = new Credis_Client(
+                isset($redisConfig['server']) ? $redisConfig['server'] : '127.0.0.1',
```
oshmyheliuk: Please use `??` ternary operator.
`$redisConfig['sever'] ?? '127.0.0.1'`

composer.json
shiftedreality: Since we support Magento `>=2.1.16`, this should be correct constraint: `^1.8.2`

https://github.com/magento/magento2ce/blob/2.1.16/composer.json#L35

src/Step/Deploy/PreDeploy/CleanRedisCache.php
```diff
@@ -74,12 +75,15 @@ public function execute()
...
+                isset($redisConfig['server']) ? $redisConfig['port'] : 6379,
```
oshmyheliuk: Why do you check on isset `$redisConfig['server']` for the `port` value?

src/Step/Deploy/PreDeploy/CleanRedisCache.php
shiftedreality: I would suggest wrapping this into simple Factory class, to avoid using `new` operator inside of a step class

Flyingmana: @shiftedreality do you have a suggestion in which namespace I should put the `CredisClientFactory`? Would `Magento\MagentoCloud\Util`be a good place?

arhiopterecs (dismissed):

composer.json
```diff
@@ -15,6 +15,7 @@
     "ext-PDO": "*",
     "ext-json": "*",
     "ext-sockets": "*",
+    "colinmollenhour/credis": "^1.8.2",
```
YPyltiai: @shiftedreality this is not correct. Magento 2.2.0 has `1.6` for example. The 2.1.x version you referenced to was updated after early 2.2.x versions. This needs to be reverted. Reference: https://github.com/magento/magento-cloud/blob/2.2.0/composer.lock#L129

YPyltiai (dismissed):

src/Step/Deploy/PreDeploy/CleanRedisCache.php
shiftedreality: Yes, that's fine

shiftedreality:
Hello @Flyingmana , could you update the PR with our suggestions?

Flyingmana:
@shiftedreality the one with wrapping it into a factory class? 
will do.

Did I miss anything else?

shiftedreality:
Looks like only this one, thanks!

shiftedreality:
Hello @Flyingmana 

I can help you and deliver missed parts to the PR, please open it for external contributors (checkbox on the right)

Flyingmana:
Thats odd, I do not have the checkbox on this PR. 

image

shiftedreality:
Yeah, that's strange. Here's the patch then: https://gist.github.com/shiftedreality/4023f8be71e79c325852fbb0981ed198

Flyingmana:
sorry for the inconvenience, and thank you for the patch file.
have added it now

shiftedreality:
Thanks @Flyingmana!

I see there is one small static error, please apply this patch as well: https://gist.github.com/shiftedreality/4ac452d91c68936b3ecf1b9f3c347709

Sorry for inconvenience

shiftedreality (dismissed):

oshmyheliuk (dismissed):

shiftedreality:
@Flyingmana could you please merge `develop` and resolve conflict? It should be pretty same

Flyingmana:
conflict was caused by the last commit being done different, when removed the merge did go without conflict and took the same change which was already added to develop.

should be fine now I think

oshmyheliuk (approved):

shiftedreality (approved):

shiftedreality:
QA approved.

Tested on Docker and Cloud instance.

PR #704, opened 2020-03-25
MCLOUD-4094: Update direct check of specific declaration in root composer.json

shiftedreality:

1. https://jira.corp.magento.com/browse/MCLOUD-4094

### Manual testing scenarios

1. Deploy Magento 2.3.4 from Composer
2. Deployment is successful
3. Deploy Magento 2.4.0 from git
4. Deployment is successful
6. Change `autoload -> psr4` from `"Zend\\Mvc\\Controller\\": "setup\/src\/Zend\/Mvc\/Controller\/"` to `"Dummy\\Mvc\\Controller\\": "setup/src/Zend/Mvc/Controller/"`
7. Deployment is failed

### Contribution checklist

### DIFF
--- a/src/Config/Validator/Build/ComposerFile.php
+++ b/src/Config/Validator/Build/ComposerFile.php
@@ -78,14 +78,25 @@ class ComposerFile implements ValidatorInterface
             return $this->resultFactory->error('Can\'t read composer.json file: ' . $e->getMessage());
         }
 
-        if (!isset($autoloadPsr4['Zend\Mvc\Controller\\'])) {
-            return $this->resultFactory->error(
-                'Required configuration is missed in autoload section of composer.json file.',
-                'Add ("Zend\\\\Mvc\\\\Controller\\\\": "setup/src/Zend/Mvc/Controller/") to autoload -> psr-4 section' .
-                ' and re-run "composer update" command locally. Then commit new composer.json and composer.lock files.'
-            );
+        $map = [
+            'Zend\Mvc\Controller\\',
+            'Laminas\Mvc\Controller\\'
+        ];
+
+        foreach ($map as $name) {
+            if (isset($autoloadPsr4[$name])) {
+                return $this->resultFactory->success();
+            }
         }
 
-        return $this->resultFactory->success();
+        return $this->resultFactory->error(
+            'Required configuration is missed in autoload section of composer.json file.',
+            sprintf(
+                'Update autoload -> psr-4  section in composer.json similar to %s '
+                . 'and re-run "composer update" command locally. '
+                . 'Then commit new composer.json and composer.lock files.',
+                'https://github.com/magento/magento-cloud/blob/master/composer.json'
+            )
+        );
     }
 }
--- a/src/Config/Validator/Build/ComposerFile.php
+++ b/src/Config/Validator/Build/ComposerFile.php
@@ -95,7 +95,7 @@ class ComposerFile implements ValidatorInterface
                 'Update autoload -> psr-4  section in composer.json similar to %s '
                 . 'and re-run "composer update" command locally. '
                 . 'Then commit new composer.json and composer.lock files.',
-                'https://github.com/magento/magento-cloud/blob/master/composer.json'
+                'https://github.com/magento/magento-cloud'
             )
         );
     }
--- a/src/Config/Validator/Build/ComposerFile.php
+++ b/src/Config/Validator/Build/ComposerFile.php
@@ -90,12 +90,11 @@ class ComposerFile implements ValidatorInterface
         }
 
         return $this->resultFactory->error(
-            'The MVC framework configuration specified in the autoload section of the `composer.json` file is missing or incorrect. Complete the following steps to fix this issue:',
+            'Required configuration is missed in autoload section of composer.json file.',
             sprintf(
-                '- Update autoload -> psr-4  section in composer.json similar to %s '
-                . 'and run the "composer update" command locally. '
-                . 'Then commit new composer.json and composer.lock files.',
-                'https://github.com/magento/magento-cloud'
+                'Add (%s) to autoload -> psr-4 section' .
+                ' and re-run "composer update" command locally. Then commit new composer.json and composer.lock files.',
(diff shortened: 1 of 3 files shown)


src/Config/Validator/Build/ComposerFile.php
```diff
@@ -78,14 +78,25 @@ public function validate(): Validator\ResultInterface
...
         }
 
-        return $this->resultFactory->success();
+        return $this->resultFactory->error(
+            'Required configuration is missed in autoload section of composer.json file.',
+            sprintf(
+                'Update autoload -> psr-4  section in composer.json similar to %s '
+                . 'and re-run "composer update" command locally. '
+                . 'Then commit new composer.json and composer.lock files.',
+                'https://github.com/magento/magento-cloud/blob/master/composer.json'
```
NadiyaS: it is not always should be example from master branch, isn't it?

shiftedreality: I changed to generic repo now.

@meker12 please review the text

src/Config/Validator/Build/ComposerFile.php
```diff
@@ -78,14 +78,25 @@ public function validate(): Validator\ResultInterface
...
         }
 
-        return $this->resultFactory->success();
+        return $this->resultFactory->error(
+            'Required configuration is missed in autoload section of composer.json file.',
+            sprintf(
+                'Update autoload -> psr-4  section in composer.json similar to %s '
+                . 'and re-run "composer update" command locally. '
+                . 'Then commit new composer.json and composer.lock files.',
+                'https://github.com/magento/magento-cloud'
```
bbatsche: This whole error message seems so vague, almost to the point of not being useful. Also, wouldn't this validator give a false negative if the user is on 2.3.4, but still has `Zend\Mvc\Controller\\` in the autoload section?

I think it would be better to modify `$map` in the form of `<constraint> => <namespace>` and then use `MagentoVersion::satisfies()` to make sure a) we have the right namespace for the right version of Magento and b) we can give a more accurate error message

bbatsche (changes requested):

src/Config/Validator/Build/ComposerFile.php
shiftedreality: There is no direct aligning between Magento and Laminas,  that's' the reason why we don't use a map here

bbatsche: I thought we were swapping out the dependency in 2.4? (or 2.3.4?) Either way, at the very least there needs to be some documentation somewhere (not here, devdocs) about "if you are using version x.x Magento use Zend, if you are using version y.y use Laminas..." If there isn't some clearly specified cutoff point, pointing to that documentation would be a lot more useful than "go to this repository, know which version/branch to pick (somehow), then look at this file in there."

Alternatively, maybe we drive this validator off of the Zend/Laminas packages instead of Magento:

1. Look for `laminas/laminas-mvc` or `zendframework/zend-mvc` packages
2. Look for correct entry in autoload section, based on which package was found
3. Report error, based on which package was found

meker12:
```suggestion
            'The MVC framework configuration specified in the autoload section of the `composer.json` file is missing or incorrect. Complete the following steps to fix this issue:',
            sprintf(
                '- Update autoload -> psr-4  section in composer.json similar to %s '
                . 'and run the "composer update" command locally. '
                . 'Then commit new composer.json and composer.lock files.',
                'https://github.com/magento/magento-cloud'
```

I don't know the syntax for the message code, but suggest output look like this:

- Update the `autoload -> psr-4`  section in the composer.json file to fix the configuration. For example,  %s '
- In the local environment, run the "composer update" command again.
- Add and commit the updated `composer.json` and `composer.lock` files.


Not sure what %s stands for.

From Ben's message, it looks like the config should be either laminas or zendframework depending on the version.  Is it possible to include that in the error message?  For x.x, use laminas.  For x.x use zend-mvc

src/Config/Validator/Build/ComposerFile.php
```diff
@@ -78,14 +78,25 @@ public function validate(): Validator\ResultInterface
...
+            }
         }
 
-        return $this->resultFactory->success();
+        return $this->resultFactory->error(
+            'The MVC framework configuration specified in the autoload section of the `composer.json` file is missing or incorrect. Complete the following steps to fix this issue:',
+            sprintf(
+                '- Update autoload -> psr-4  section in composer.json similar to %s '
+                . 'and run the "composer update" command locally. '
+                . 'Then commit new composer.json and composer.lock files.',
```
meker12:
```suggestion
                . 'Then, commit new composer.json and composer.lock files.',
```

src/Config/Validator/Build/ComposerFile.php
```diff
@@ -78,14 +78,25 @@ public function validate(): Validator\ResultInterface
...
+            if (isset($autoloadPsr4[$name])) {
+                return $this->resultFactory->success();
+            }
         }
 
-        return $this->resultFactory->success();
+        return $this->resultFactory->error(
+            'The MVC framework configuration specified in the autoload section of the `composer.json` file is missing or incorrect. Complete the following steps to fix this issue:',
+            sprintf(
+                '- Update autoload -> psr-4  section in composer.json similar to %s '
```
meker12:
```suggestion
                '- Add the configuration to the autoload -> psr-4  section in the composer.json similar to %s '
```

src/Config/Validator/Build/ComposerFile.php
shiftedreality: I just realized that technically there is only one solution now - `Laminas`, so I changed suggestion to point to `Laminas` but keep a check for both frameworks.

shiftedreality: But it seems like it applies to only newer versions of the framework, so I'll apply your suggestion here

bbatsche (approved):

NadiyaS (approved):

mveeramneni:
QA Approved

PR #706, opened 2020-03-27
MCLOUD-5325: Move log section from .magento.env.yaml

oshmyheliuk:
### Description
https://jira.corp.magento.com/browse/MCLOUD-5325

### Manual testing scenarios

1. Clone magento-cloud project locally
2. Run command `vendor/bin/ece-tools schema:generate`
3. Check that the generated file contains content from `vendor/magento/ece-tools/dist/.log.env.md` at the end of the file.

### Contribution checklist


src/Filesystem/FileList.php
```diff
@@ -132,6 +132,14 @@ public function getEnvDistConfig(): string
         return $this->directoryList->getMagentoRoot() . '/.magento.env.md';
     }
 
+    /**
+     * @return string
+     */
+    public function getLogDistConfig(): string
```
shiftedreality, line 138: The naming should be aligned with the main config file.

Something like `.magento-log.env.yaml`

@YPyltiai

src/Filesystem/FileList.php
```diff
@@ -132,6 +132,14 @@ public function getEnvDistConfig(): string
...
+    {
+        return $this->directoryList->getRoot() . '/dist/.log.env.md';
```
shiftedreality, line 140: This line is not covered with Unit Test

shiftedreality (changes requested):

src/Filesystem/FileList.php
oshmyheliuk: added

src/Filesystem/FileList.php
YPyltiai: @shiftedreality if this is just a file to contain text to merge with .magento.env.yaml - why would we care about name? It's technically not moving functionality from .magento.env.yaml AFAIK.

oshmyheliuk: Yes, it's just a file for storing log markdown, it won't be copied into the project root, only its content will be added to generated .magento.env.md file.

dist/.log.env.md
```diff
@@ -0,0 +1,90 @@
+### Set up notifications
```
bbatsche: The organization of this file doesn't make a lot of sense to me. Slack & Email configuration are log handlers and should go under there (that would mean modifying some of the language as well). I think we should re-order to the following:
```
# Logging Handlers
## Slack Configuration
## Email Configuration
# Log Levels
```

bbatsche (changes requested):
If we feel this isn't worth worrying about right now I could change this to an approval, but if we're modifying this file anyway it would be good to clean it up a bit

dist/.log.env.md
oshmyheliuk: Also, on devdocs we have two pages about the logging configuration.
https://devdocs.magento.com/cloud/env/log-handlers.html
https://devdocs.magento.com/cloud/env/setup-notifications.html

@meker12 could you please take a look at these pages and this file, we need to combine it somehow better.

NadiyaS (dismissed):

dist/.log.env.md
```diff
@@ -0,0 +1,93 @@
+
```
bbatsche: Remove this extra blank line here. If we need more space between the schema and this info we should add it in `GenerateSchema.php`.

dist/.log.env.md
```diff
@@ -0,0 +1,93 @@
...
+### Logging handlers
```
bbatsche: It doesn't make sense to have this many `#` at the start of the file; it's the equivalent of starting a HTML page with an `h3`. Hierarchy should be:
```
# Logging handlers
## Send logs to a remote logging server
## Send logs to a messaging system
# Log levels
```

bbatsche (changes requested):
Two small changes, restructure looks really good though

dist/.log.env.md
oshmyheliuk: I'm not sure about this, it was PR from @meker12 https://github.com/magento/ece-tools/pull/709/files
@meker12 Could you please check comment from Ben

meker12: So I was assuming that this information was being appended to another file -- so I followed the hierarchy that Sasha used, but your comment makes sense @Ben. Need to check the schema file.

oshmyheliuk: The content of this file will be a part of automatically generated `.md` file and there is already used `##` and `###`, so it will be strange if after `##` will be added header `#`.
Part of the file
```
## CRON_CONSUMERS_RUNNER

Use this variable to make sure message queues are running after a deployment. By default, the deployment process
overwrites all settings in the env.php file

 cron_run — a boolean value that enables or disables the consumers_runner cron job.

 max_messages — a number specifying the maximum number of messages each consumer must process before terminating.
Although we do not recommend it, you can use 0 to prevent the consumer from terminating.

 consumers — an array of strings specifying which consumer(s) to run. An empty array runs all consumers.

|Attribute|Values|
|---|---|
|Type|array|
|Magento version|\>=2.1.14|
|Stages|deploy|

### Examples

```yaml
  stage:
      deploy:
          CRON_CONSUMERS_RUNNER:
              cron_run: true
              max_messages: 1000
              consumers:
                  - consumer1
                  - consumer2

```

oshmyheliuk: Looks like we need to change to `###` -> ` ##`  and `####` -> `###`

meker12: so I think Ben is correct that the levels are not quite correct.

I believe that the variable descriptions need to be at the H3 level, and the titles like Logging handlers will be at the h2 level.

This is setting the markdown up to be inserted into a template for a doc page-- 

 
image

dist/.log.env.md
oshmyheliuk: removed

NadiyaS (dismissed):

bbatsche (dismissed):

bbatsche (approved):

NadiyaS (approved):

shiftedreality (approved):

mveeramneni:
QA approved

ghost:
Hi @oshmyheliuk, thank you for your contribution! </br>Please, complete [Contribution Survey](https://opensource.magento.com/survey/contribution/magento+ece-tools+706), it will take less than a minute. </br>Your feedback will help us to improve contribution process.

PR #722, opened 2020-04-23
Issue 667 : Extend environment data to be read from local file

### LINKED ISSUE #667: post_deploy hook not correctly retrieved
When you check the configuration with `vendor/bin/ece-tools wizard:ideal-state` the output is always `Your application does not have the "post_deploy" hook enabled.` In fact the hook is correctly configured in .magento.app.yaml. But the check in src/Config/Application/HookChecker.php reads the setting from a environment variable (ENV_APPLICATION) rather than from .magento.app.yaml. Instead Magento\MagentoCloud\Filesystem\FileList::getAppConfig() should get called to retrieve the setting from the file.

Is my assumption correct?

BarnyShergold:
### Description
Updated getApplication to try and read physical .magento.app.yaml if ENV is empty - this happens on local instances.
This will allow higher level uses to work on local installations. Specifically this was done to allow `vendor/bin/ece-tools wizard:ideal-state` to report correctly

1. magento/ece-tools#667: post_deploy hook not correctly retrieved

### Manual testing scenarios

1. on a local instance run 'vendor/bin/ece-tools wizard:ideal-state' - if .magento.app.yaml is present and the "post_deploy" hook correctly configured then the return will be that the application is in an ideal state
2. Also running locally and REMOVING the file or having post_deploy hook not configured will cause normal warning message. Missing file will not cause an  untrapped exception.

### Contribution checklist


src/Config/EnvironmentData.php
```diff
@@ -114,18 +136,31 @@ public function getApplication(): array
             return $this->data['application'];
         }
 
-        return $this->data['application'] = $this->getEnvVar(SystemConfigInterface::VAR_ENV_APPLICATION, []);
+        $application = $this->getEnvVar(SystemConfigInterface::VAR_ENV_APPLICATION, []);
+
+        if (!$application) {
+            // Try to read from .magento.app.yaml
```
oshmyheliuk: I would recommend removing comments as it's clear what code is doing.

src/Config/EnvironmentData.php
```diff
@@ -114,18 +136,31 @@ public function getApplication(): array
...
+            try {
+                $application = Yaml::parse(
+                    $this->file->fileGetContents($this->fileList->getAppConfig())
+                );
+            } catch (FileSystemException $e) {
+                // Do nothing as $application needs to be empty
```
oshmyheliuk, line 147: Instead of a comment, I think better to add some warning message about neither env variable and .magento.app.yaml  not exists

src/Config/EnvironmentData.php
```diff
@@ -114,18 +136,31 @@ public function getApplication(): array
...
+     *
      */
     public function getBranchName(): string
     {
         $envVarName = $this->systemConfig->get(SystemConfigInterface::VAR_ENV_ENVIRONMENT);
 
         return $this->getEnv($envVarName) ? (string) $this->getEnv($envVarName) : '';
     }
-}
+}
```
oshmyheliuk: Please add new line

src/Config/EnvironmentData.php
oshmyheliuk: But I'm not sure about the message, @YPyltiai what do you think?

BarnyShergold: Well, I left it as nothing as that is the current behaviour. I did think about putting a message in but left it. If the consensus is that at message is needed, should it be output direct to console line or done as a log? I'd lean towards the former as in theory it should only happen on a local instance if the file is missing.

oshmyheliuk: I've discussed with @YPyltiai and we think that message is not needed. Let's leave as it is, thank you.

BarnyShergold: Okay - I made the other changes you requested so can you approve the PR?
BTW I don't understand why the CI Checks are failing. The failure doesn't seem appropriate ?
@oshmyheliuk @YPyltiai

BarnyShergold:
@bbatsche @oshmyheliuk @YPyltiai  - can we not get approval on this PR please? It had minor corrections made 4 days ago and we could get this into QA and moved on? Please? :)

src/Config/EnvironmentData.php
oshmyheliuk: Yes, sorry, I thought that I've already approved

oshmyheliuk (dismissed):

shiftedreality:
Hello @BarnyShergold,

This issue is fixed for the Cloud Docker installation. Could you provide another example of a local installation?

BarnyShergold:
@shiftedreality - Yes - currently Vaimo developers use Vagrant. This is what I tested it against.

bbatsche (dismissed):

shiftedreality:
@BarnyShergold could you please fix CI tests?

shiftedreality:
There is an issue with secrets for community PRs, I'll need to fix it first to unblock this PR

BarnyShergold:
@bbatsche  @oshmyheliuk @shiftedreality - All CI Checks have now passed (after SR fixed the CI a couple of small code formatting issues arose, which I fixed) and Travis has passed

Can you please re-review and then we can this PR moved on :)

oshmyheliuk (approved):

bbatsche (approved):

andriyShevtsov:
QA approved

oshmyheliuk:
@BarnyShergold Thank you for your contribution.

PR #726, opened 2020-05-05
MCLOUD-5637: Update all error codes in ece-tools

oshmyheliuk:
### Description
https://jira.corp.magento.com/browse/MCLOUD-5637

### Fixed Issues (if relevant)

https://jira.corp.magento.com/browse/MCLOUD-5637

### Manual testing scenarios

1. Deploy Magento with the wrong configuration in `.magento.env.yaml`
```yaml
stage:
    deploy:
        SCD_THREADS: "asd"
```
2. Check that the deployment process failed with error code different than 1.
```
[2020-05-22 14:45:01] ERROR: [10] Fix configuration with given suggestions:
        - Environment configuration is not valid. Correct the following items in your .magento.env.yaml file:
          The SCD_THREADS variable contains an invalid value of type string. Use the following type: integer.
````
3. Fix `.magento.env.yaml`  and add some wrong configuration through UI. For example 
`DATABASE_CONFIGURATION` => `{'key':'value'}`
4. Check that the deployment process finished with error code different than 1.
```
    [2020-05-22 13:52:57] ERROR: [109] Fix configuration with given suggestions:
    - Variable DATABASE_CONFIGURATION is not configured properly
      At least host, dbname, username and password options must be configured for default connection  
```
### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #732, opened 2020-05-20
MCLOUD-5991: Restore additional data in env.php if this file was removed

NadiyaS:
### Description
If env.php does not exist and DB is full of data, ece-tools runs setup:upgrade, but the deployment is still failed. 

- check existence of crypt key. If this data does not exist (whether in env.php or in .magento.env.yaml) - throw exception that there is not enough data for setup:upgrade or DB is not empty for setup:install and provide appropriate steps to resolve this. Otherwise, continue to run Magento upgrade process as it exists now.
- if env.php did not exist or had only one cache type in it - restore all default types, otherwise keep them as they are (we cannot restore cache type data by default as customer could remove some types for purpose)
- add warning that env.php did not exist at the beginning of deployment.

1. https://jira.corp.magento.com/browse/MCLOUD-5991

### Manual testing scenarios

1. on installed Magento remove env.php and run redeployment 
**Expected result:** deployment is failed with message that `crypt key` does not exist
2. remove env.php again (as it was partially restored form prev. deployment) add CRYPT_KEY variable and run redeployment
**Expected result:**
- deployment finished without errors
- env.php has whole list of `cache_types`
- cloud.log contain message that cache_types was restored and that env.php did not exist at the beginning of the deployment
3. in env.php remove a few types in `cache_type` option and set `0` to other cache_types and left few with value `1`. Run redeployment.
**Expected result:** deployment finished. env.php has the same state of `cache_types` option (few removed types, and types with values as 0 and 1)

### Contribution checklist


oshmyheliuk (dismissed):

oshmyheliuk (dismissed):

oshmyheliuk (dismissed):

src/Filesystem/Flag/Manager.php
```diff
@@ -37,6 +37,11 @@ class Manager
      */
     public const FLAG_IGNORE_SPLIT_DB = 'ignore_split_db';
 
+    /**
+     * Used to mark that env.php file does not exist at the beginning of deployment
+     */
+    public const FLAG_ENV_FILE_ABSENCE = 'env_file_absence';
```
bbatsche, line 43: My understanding was that the flags were useful to communicate between build & deploy stages, or between multiple deployments. Why are we using them here were everything is happening within a single deployment?

src/Step/Deploy/InstallUpdate/ConfigUpdate/CacheType.php
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            if ($this->flagManager->exists(FlagManager::FLAG_ENV_FILE_ABSENCE)) {
+                $this->logger->info('Run cache:enable to restore all cache types');
+                $this->magentoShell->execute(
+                    'cache:enable',
+                    [$this->stageConfig->get(DeployInterface::VAR_VERBOSE_COMMANDS)]
+                );
+            }
+        } catch (ShellException $exception) {
+            //code will be changed after merging PR with error reporting
```
bbatsche: If I understand your comment correctly, this exception will change slightly once specific error codes are created? Otherwise it seems like we should be able to just catch `GenericException`s below.

bbatsche (changes requested):
I have a couple of questions but ultimately the code looks good. I'll approve once they're answered.

src/Step/Deploy/InstallUpdate/ConfigUpdate/CacheType.php
NadiyaS: yes, as the codes for theres exceptions will be different I caught both types, just not to forget about it after merging PR https://github.com/magento/ece-tools/pull/726/, e.g. https://github.com/magento/ece-tools/pull/726/files#diff-11ebfae7dd9722de4d90062c73b5170fR89

src/Filesystem/Flag/Manager.php
NadiyaS: good question. Actually, we can understand whether env.php was removed or not only at the beginning of deploy phase, because already on the second step `cache` https://github.com/magento/ece-tools/blob/develop/scenario/deploy.xml#L8 we write data to env.php file and if it is absent we create this file with some part of data. But steps which should react on file absence go much later and at that moment file always exist

bbatsche (dismissed):

oshmyheliuk (dismissed):

oshmyheliuk (dismissed):

bbatsche (dismissed):

oshmyheliuk (approved):

bbatsche (approved):

andriyShevtsov:
QA approved

